PR #15, opened 2026-03-09
Match Linux terminal copy/paste behavior

joshdoe:
## Summary
- align terminal copy/paste behavior on Linux with the expected platform flow
- left-drag selects, right-click when selection is present clears the selection, right-click with no selection pastes
- wire the Electron and preload IPC needed for the updated terminal clipboard path
- update the terminal view to use the platform-specific handling

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -58,6 +58,8 @@ export enum IPC {
   // Dialog
   DialogConfirm = '__dialog_confirm',
   DialogOpen = '__dialog_open',
+  ClipboardRead = '__clipboard_read',
+  ClipboardWrite = '__clipboard_write',
 
   // Shell
   ShellReveal = '__shell_reveal',
--- a/electron/ipc/pty.test.ts
+++ b/electron/ipc/pty.test.ts
@@ -21,10 +21,6 @@ describe('validateCommand', () => {
     );
   });
 
-  it('does not throw for a bare command found in PATH', () => {
-    expect(() => validateCommand('sh')).not.toThrow();
-  });
-
   it('throws for an empty command string', () => {
     expect(() => validateCommand('')).toThrow(/must not be empty/);
   });
--- a/electron/ipc/register.test.ts
+++ b/electron/ipc/register.test.ts
@@ -1,0 +1,161 @@
+import { beforeEach, afterEach, describe, expect, it, vi } from 'vitest';
+import { IPC } from './channels.js';
+
+type IpcHandler = (event: unknown, args?: unknown) => unknown;
+
+const mockState = vi.hoisted(() => ({
+  handlers: new Map<string, IpcHandler>(),
+  clipboardReadText: vi.fn(),
+  clipboardWriteText: vi.fn(),
+}));
+
+vi.mock('electron', () => ({
+  ipcMain: {
+    handle: vi.fn((channel: string, handler: IpcHandler) => {
+      mockState.handlers.set(channel, handler);
+    }),
+  },
+  dialog: {},
+  shell: {},
+  app: {
+    getPath: vi.fn(() => '/tmp'),
+  },
+  BrowserWindow: function BrowserWindow() {},
+  clipboard: {
+    readText: mockState.clipboardReadText,
+    writeText: mockState.clipboardWriteText,
+  },
+}));
+
+vi.mock('./pty.js', () => ({
+  spawnAgent: vi.fn(),
+  writeToAgent: vi.fn(),
+  resizeAgent: vi.fn(),
+  pauseAgent: vi.fn(),
+  resumeAgent: vi.fn(),
+  killAgent: vi.fn(),
+  countRunningAgents: vi.fn(),
+  killAllAgents: vi.fn(),
+  getAgentMeta: vi.fn(),
+}));
+
+vi.mock('./plans.js', () => ({
+  ensurePlansDirectory: vi.fn(),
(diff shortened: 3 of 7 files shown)


johannesjo (commented):
## Code Review

**Overall: Well-structured PR that correctly solves a real Linux clipboard limitation.** Routing clipboard through Electron IPC is the right call — browser `navigator.clipboard` can't access the X11 selection buffer. Clean implementation, proper cleanup, good test coverage on the backend.

---

### Important

**1. Needs rebase — merge conflict on `register.ts` line 1**
Main now imports `Notification` from electron (from `587fde6`). This PR changes the same line, so it will conflict. Resolved import should be:
```ts
import { ipcMain, dialog, shell, app, BrowserWindow, Notification, clipboard } from 'electron';
```

**2. Incomplete test mocks in `register.test.ts`**
After rebase, the electron mock will need `Notification`. Also missing from mocks: `getAllFileDiffs`/`getAllFileDiffsFromBranch` in `git.js`, and `plans.js` exports `stopAllPlanWatchers` but `register.ts` imports `stopPlanWatcher` (singular) + `readPlanForWorktree`.

**3. `pasteClipboard` silently swallows errors**
```ts
.catch(() => {});
```
A `console.warn` would help debug clipboard failures across Linux DEs (GNOME, KDE, Wayland, X11).

**4. Unrelated test removal in `pty.test.ts`**
The "bare command found in PATH" test deletion is unrelated to clipboard work — should be a separate commit.

**5. `ClipboardWrite` with `target: 'selection'` on non-Linux writes nothing**
```ts
if (target !== 'selection') clipboard.writeText(args.text);  // skipped
if ((target === 'selection' || target === 'both') && process.platform === 'linux') {  // skipped on macOS
```
Dead path today (frontend only sends `'both'` or `'clipboard'`), but worth either falling back to the default clipboard or throwing on unsupported platform+target combos.

---

### Minor

- The `requestAnimationFrame` in `handleContextMenu` could use a brief comment explaining why the deferral is needed (letting browser finalize selection after right-click).
- Missing test cases: `ClipboardWrite` with explicit `'clipboard'` target, `ClipboardRead` with no target, `ClipboardWrite` with `'selection'` on non-Linux.

johannesjo:
Thank you very much! <3

johannesjo:
## Review — additional findings after rebase consideration

The existing review covers the main blockers well. Two additional issues after comparing the PR's diff against the current `main`:

### 1. Image paste regression after rebase (needs attention during conflict resolution)

On current `main`, the `isPaste` keyboard handler has a two-stage fallback:
```ts
const text = await navigator.clipboard.readText().catch(() => '');
if (text) { enqueueInput(text); return; }
// Fall back to clipboard image → save to temp file and paste path
const filePath = await invoke<string | null>(IPC.SaveClipboardImage);
if (filePath) enqueueInput(filePath);
```

The PR replaces this entire block with `pasteClipboard()` (which only reads text via IPC). The PR's diff base is an older version of `main` that did not have this image fallback. When rebased, whoever resolves the conflict must preserve the `IPC.SaveClipboardImage` fallback for non-image text paste paths, otherwise Ctrl+Shift+V with an image in the clipboard will silently break on Linux.

Suggested approach after rebase — keep the image fallback in the Linux path:
```ts
if (isPaste) {
  pasteClipboard(); // reads from clipboard (text) via IPC
  // still preserve image fallback if text is empty? or accept the drop?
  return false;
}
```
At minimum this should be an explicit decision, not an accidental omission from a merge conflict.

### 2. `pasteClipboard` silently swallows errors — already noted, but the contrast is worth highlighting

`copySelectionToClipboard` uses `fireAndForget` (which logs errors to console), while `pasteClipboard` uses `.catch(() => {})` (completely silent). Both should use the same error strategy. On Wayland/KDE, `xclip`/`xdotool` access failures are common and hard to debug without at least a `console.warn`.

johannesjo (changes requested):
The PR is currently unmergeable due to upstream conflicts. More importantly, after a rebase a regression will be introduced: a `SaveClipboardImage` fallback was added to the `isPaste` keyboard handler on `main` since this branch was cut — the PR's replacement of that block with `pasteClipboard()` will silently drop it, breaking Ctrl+Shift+V with images in the clipboard on Linux.

Please: (1) rebase onto main, (2) port the image-paste fallback into `pasteClipboard()`, and (3) address the existing review findings (import conflict, broken test mocks, silent error swallowing, unrelated test deletion). Details in the earlier review comment.

johannesjo:
Thanks again for putting this together, @joshdoe — and apologies for the slow back-and-forth.

  After taking another look, I'm going to close this one out. The basic Ctrl+Shift+C/V copy-paste path on Linux already works on main, and the rest of the PR mixes a few different conventions:

  - Middle-click paste from the X11 PRIMARY selection — this is the genuinely useful piece, and the only thing navigator.clipboard really can't do. I'd like to land a focused change just for this.
  - Right-click → copy/paste — this is PuTTY/WSL behavior; gnome-terminal/konsole/xterm all show a context menu instead, so I don't want to override it by default.
  - Ctrl+Insert / Shift+Insert — happy to add if there's demand, but I haven't seen it requested.

  Rather than rebase the combined PR (which also needs the new SaveClipboardImage fallback ported into the paste path to avoid a regression), I'll open a small issue tracking just the X11 PRIMARY / middle-click bit. If you'd like to
  take another swing at that narrower scope, I'd love to merge it — otherwise no worries, I'll pick it up. Either way, much appreciated for surfacing this. ❤️

PR #17, opened 2026-03-09
Fix task creation reload in dev

joshdoe:
## Summary
- preserve task creation behavior during Electron dev reloads
- include the needed dev config handling in the main process and Electron Vite setup

## Testing
- not run here

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -51,7 +51,9 @@ function verifyPreloadAllowlist(): void {
     const preloadPath = path.join(__dirname, '..', 'electron', 'preload.cjs');
     const preloadSrc = fs.readFileSync(preloadPath, 'utf8');
     const enumValues = new Set(Object.values(IPC));
-    const missing = [...enumValues].filter((v) => !preloadSrc.includes(`"${v}"`));
+    const hasChannel = (channel: string) =>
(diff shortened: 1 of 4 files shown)


johannesjo (commented):
## Review Feedback

Overall this looks good and safe to merge. Three small, well-scoped fixes. One item worth discussing:

### `await saveState()` — consider fire-and-forget

`await saveState()` in both `createTask` and `createDirectTask` adds disk I/O latency to the task creation return path. Since `saveState()` already has its own `.catch()` handler internally, calling it without `await` would give the same persistence guarantee without blocking the caller:

```ts
saveState(); // fire-and-forget — errors handled internally
```

The only reason to keep `await` would be if you need the write fully flushed before returning, but a Vite dev reload (triggered by a file watcher in a separate process) is extremely unlikely to race with the same event loop tick.

### Minor: PR description doesn't mention the preload quote fix

The `hasChannel` fix in `main.ts` (matching single-quoted channels in `preload.cjs`) is a separate bug fix from the reload/persistence changes. Worth calling out in the description or splitting into its own commit for cleaner history.

### Nitpick: `siblingWorktreesDir` naming

This variable actually resolves to the parent directory, not specifically a worktrees directory. `parentDir` would be more precise, though the intent is clear from context.

johannesjo:
Thank you very much! <3

PR #43, opened 2026-03-28
feat(terminal): add clickable file path links

maskar:
## Summary
File paths in terminal output are now highlighted and clickable. Clicking opens the file with the system default handler.

Supports:
- Absolute paths: `/Users/foo/bar.ts`
- Relative paths: `./src/lib/theme.ts`, `../package.json`
- Bare relative: `src/components/Foo.tsx`
- Scoped packages: `@xterm/xterm/typings/xterm.d.ts`
- Line/col suffixes: `src/App.tsx:42:10` (stripped before opening)
- Trailing punctuation is stripped (e.g. `file.md.` → `file.md`)

Also adds an `onFileLink` callback prop to `TerminalView` for parent components to handle specific file types differently.

## Test plan
- [ ] File paths in terminal output are underlined on hover
- [ ] Clicking a file path opens it in the default application
- [ ] Paths with `:line:col` suffixes are recognized and opened correctly
- [ ] Trailing punctuation (`.`, `,`, `)`) is not included in the link

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -98,6 +98,10 @@ export enum IPC {
   // System
   GetSystemFonts = 'get_system_fonts',
 
+  // File links
+  OpenPath = 'open_path',
+  ReadFileText = 'read_file_text',
+
   // Notifications
   ShowNotification = 'show_notification',
   NotificationClicked = 'notification_clicked',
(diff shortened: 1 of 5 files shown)


johannesjo:
Want me to open a fresh PR from the same branch? Yes, that would be great! Though I think I might have somehow merged this. Maybe together with one of the other PRs?

PR #53, opened 2026-04-03
feat: add MiniMax M2.7 as configurable backend for Ask about Code

octo-patch:
## Summary

- Adds **MiniMax M2.7** (204K context) as an alternative LLM provider for the _Ask about Code_ inline Q&A feature
- Users can switch between Claude Code (default) and MiniMax in the **Settings → Ask about Code** section
- When MiniMax is selected, the app calls the MiniMax API directly via its OpenAI-compatible endpoint — no `claude` CLI required
- API key is stored in app state and persisted across restarts

## Changes

| File | What changed |
|------|-------------|
| `electron/ipc/ask-code-minimax.ts` | New MiniMax streaming backend: SSE parser, abort-signal handling, Bearer auth |
| `electron/ipc/ask-code-minimax.test.ts` | 12 unit tests (chunks, errors, cancel, auth header, model, temperature, stream flag) |
| `electron/ipc/ask-code.ts` | Route to MiniMax or Claude based on `provider` arg |
| `electron/ipc/register.ts` | Forward `provider` / `minimaxApiKey` from IPC call args |
| `src/store/{types,core,ui,store,persistence}.ts` | New `askCodeProvider` + `minimaxApiKey` store fields, persisted to state.json |
| `src/components/SettingsDialog.tsx` | Provider dropdown + conditional API-key input |
| `src/components/AskCodeCard.tsx` | Pass provider + apiKey through IPC |
| `README.md` | Mention MiniMax as a supported Ask-about-Code provider |

## Test plan

- [ ] Run `npm test` — 12 new passing tests, all existing tests still pass
- [ ] Open Settings (`Ctrl+,`) → see new "Ask about Code" section with provider dropdown
- [ ] Select MiniMax, enter a valid `MINIMAX_API_KEY`, select code in a task, ask a question — response streams in
- [ ] Switch back to Claude Code — Claude CLI is used as before
- [ ] Restart the app — provider and key are persisted

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@
 - **See every session in one place** — switch context without losing momentum.
 - **Control everything keyboard-first** — every action has a shortcut, mouse optional.
 - **Monitor progress from your phone** — scan a QR code, watch agents work over Wi-Fi or Tailscale.
+- **Ask about code with any LLM** — the inline code Q&A feature supports [Claude Code](https://docs.anthropic.com/en/docs/claude-code) (default) or [MiniMax](https://www.minimax.io/) M2.7 (204K context) — configurable in Settings.
 
 <details>
 <summary><strong>How does it compare?</strong></summary>
--- a/electron/ipc/ask-code-minimax.test.ts
+++ b/electron/ipc/ask-code-minimax.test.ts
@@ -4,7 +4,12 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 const mockFetch = vi.fn();
 vi.stubGlobal('fetch', mockFetch);
 
-import { askAboutCodeMinimax, cancelAskAboutCodeMinimax, MINIMAX_MODEL } from './ask-code-minimax.js';
+import {
+  askAboutCodeMinimax,
+  cancelAskAboutCodeMinimax,
+  MINIMAX_MODEL,
+  setMinimaxApiKey,
+} from './ask-code-minimax.js';
 
 function makeMockWin() {
   const messages: unknown[] = [];
@@ -57,6 +62,7 @@ function sseChunk(content: string): string {
 describe('askAboutCodeMinimax', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    setMinimaxApiKey('test-key');
   });
 
   it('throws if prompt exceeds max length', () => {
@@ -67,7 +73,6 @@ describe('askAboutCodeMinimax', () => {
         requestId: 'r1',
         channelId: 'ch1',
         prompt: longPrompt,
-        apiKey: 'test-key',
       }),
     ).toThrow(/Prompt too long/);
   });
@@ -82,7 +87,6 @@ describe('askAboutCodeMinimax', () => {
       requestId: 'r2',
       channelId: 'ch2',
       prompt: 'Explain this code',
-      apiKey: 'test-key',
     });
 
     await waitForDone(messages);
@@ -106,7 +110,6 @@ describe('askAboutCodeMinimax', () => {
       requestId: 'r3',
       channelId: 'ch3',
       prompt: 'What is this?',
-      apiKey: 'bad-key',
     });
 
     await waitForDone(messages);
@@ -128,7 +131,6 @@ describe('askAboutCodeMinimax', () => {
       requestId: 'r4',
       channelId: 'ch4',
       prompt: 'Explain',
-      apiKey: 'test-key',
     });
 
     await waitForDone(messages);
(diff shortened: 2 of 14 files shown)


johannesjo:
## Review findings (new issues not covered by Copilot)

### 1. API key flows through IPC and lives in renderer store (medium severity)

In `AskCodeCard.tsx`, `store.minimaxApiKey` is read from the SolidJS renderer store and passed directly as an IPC argument on every query:

```ts
minimaxApiKey: store.minimaxApiKey,
```

This means:
- The key lives in plaintext renderer-process memory (accessible to any compromised renderer script or DevTools).
- The key is serialised into every IPC `AskAboutCode` message, so it appears in Electron's IPC log when devTools IPC logging is enabled.

Copilot's comment covered plaintext persistence to `state.json`; this is the complementary concern about the in-flight IPC surface. A safer architecture is to keep the key **only in the main process** (store it there after the user types it once via a dedicated `set_minimax_api_key` IPC call, ideally via `safeStorage`) and never send it back to the renderer or forward it per-request.

---

### 2. User-initiated cancel sends `exitCode: 1` (semantic bug, low severity)

In `ask-code-minimax.ts`, both the timeout path and the abort path emit `{ type: 'done', exitCode: 1 }`. For a user-initiated cancel the AbortError is correctly suppressed from the error message, but the `exitCode: 1` still fires. If any caller treats exit code 1 as a failure (e.g. showing a red error state), users will see a spurious failure indicator after every cancel. Consider a distinct `exitCode: 2` or a `type: 'cancelled'` message for graceful cancellation.

---

### 3. `cancelAskAboutCode` unconditionally calls `cancelAskAboutCodeMinimax` (minor, architectural concern)

```ts
export function cancelAskAboutCode(requestId: string): void {
+  cancelAskAboutCodeMinimax(requestId);  // always called, even for claude provider
  const proc = activeRequests.get(requestId);
  ...
}
```

Today this is a no-op for claude requests (the ID won't exist in the minimax map), but it couples the cancel dispatcher to every provider unconditionally. If a third provider is added the same pattern will continue to diverge. A cleaner approach is to store the active provider per `requestId` and dispatch cancel accordingly.

---

### 4. No `max_tokens` cap in the request body (medium severity)

The MiniMax request body has no `max_tokens` field. M2.7 has a 204K context window — without a cap every response can consume up to the full output token budget. For a code Q&A use-case this risks unexpected cost and latency spikes. Adding a reasonable cap (e.g. `max_tokens: 2048`) would bound both.

---

**Summary:** Issues 1 and 4 are the most actionable before merge. Issue 2 is a UX correctness bug. Issue 3 is a code-quality concern but not a blocker.

johannesjo:
Thank you very much! <3

johannesjo (changes requested):
Three issues to address before merge:

1. **API key exposure** — the MiniMax API key is held in plaintext renderer memory and transmitted in the IPC payload on every query. Consider keeping it in the main process only, reading it there from persisted state.
2. **No `max_tokens` cap** — without a limit, a single query can produce an unbounded response, leading to unexpected latency and cost.
3. **Cancel exit code** — user-initiated cancellation emits `exitCode: 1` (treated as an error); it should emit a neutral/cancelled signal instead.

Additionally, `cancelAskAboutCode` calls the MiniMax canceller regardless of which provider is active — it should be guarded by a provider check. Details in the review comment.

octo-patch:
Thanks for the detailed review! All four issues plus the Copilot feedback have been addressed in commit dfb2e00:

**1. API key exposure (medium):**
- Added a dedicated `SetMinimaxApiKey` IPC channel (`channels.ts`, `preload.cjs`)
- API key is now stored only in main-process memory (`ask-code-minimax.ts`: `storedApiKey` + `setMinimaxApiKey()`)
- Removed `apiKey` from `MinimaxAskCodeRequest` interface — `askAboutCodeMinimax()` reads from main-process state
- Removed `minimaxApiKey` from `AskCodeRequest` in `ask-code.ts` and from the IPC args in `register.ts`
- `AskCodeCard.tsx` no longer sends `minimaxApiKey` in the IPC payload
- `persistence.ts` no longer persists the API key to `state.json`
- `SettingsDialog.tsx` input sends the key to main process via IPC on input; no longer binds to store value
- Removed `minimaxApiKey` from `AppStore` interface and store defaults

**2. No max_tokens cap (medium):**
- Added `max_tokens: 2048` to the MiniMax request body in `ask-code-minimax.ts`

**3. Cancel exit code (low):**
- Abort/cancel path now emits `{ type: 'done', exitCode: 0, cancelled: true }` instead of `exitCode: 1`
- Both the stream-abort path (`.then` branch) and the `AbortError` catch path use `exitCode: 0`

**4. cancelAskAboutCode provider guard:**
- Added `activeProviders` map in `ask-code.ts` to track which provider each request uses
- `cancelAskAboutCode()` now checks the provider and only calls `cancelAskAboutCodeMinimax()` when `provider === 'minimax'`

**Copilot feedback:**
- Removed unused `rejectReader` variable in `ask-code-minimax.test.ts`
- Narrowed `PersistedState.askCodeProvider` from `string` to `'claude' | 'minimax'`

All 12 unit tests pass. TypeScript typecheck passes with no errors.

johannesjo:
Thank you very much!

PR #58, opened 2026-04-05
fix: resolve shell tabs via OS user shell

ASRagab:
## Summary
- resolve default shell selection via `os.userInfo().shell` before falling back to `process.env.SHELL`
- use the shared resolver for both PTY shell tabs and PATH bootstrap in the Electron main process
- add focused tests covering OS shell, env fallback, and `/bin/sh` fallback behavior

## Why
Parallel Code currently falls back to the inherited `SHELL` environment variable when no explicit shell command is provided. That can cause shell tabs to launch `bash` even when the user's actual account login shell is `zsh` (or another shell), depending on how Electron was launched.

Using the OS-reported user shell first makes the default shell selection more faithful and predictable.

## Test Plan
- `npm run typecheck`
- `npm test`

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -12,5 +12,3 @@ package-lock.json
 *.AppImage
 *.deb
 *.dmg
-CLAUDE.local.md
-.enzyme/
--- a/.prettierignore
+++ b/.prettierignore
@@ -12,3 +12,5 @@ package-lock.json
 *.AppImage
 *.deb
 *.dmg
+CLAUDE.local.md
+.enzyme/
--- a/electron/ipc/pty.ts
+++ b/electron/ipc/pty.ts
@@ -6,6 +6,7 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 import type { BrowserWindow } from 'electron';
 import { RingBuffer } from '../remote/ring-buffer.js';
+import { resolveUserShell } from '../user-shell.js';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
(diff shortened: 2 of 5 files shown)


ASRagab:
@johannesjo welcome back. Updated.

johannesjo:
**Review notes** (Copilot's two comments appear already addressed in the current version — `canUseShell`/`isExecutablePosixShell` added, import uses `'os'` not `'node:os'`.)

Two new items worth considering:

**1. Unrelated `.prettierignore` changes**

The diff includes two lines added to `.prettierignore`:
```
CLAUDE.local.md
.enzyme/
```
These appear to be local developer config that crept in. They have no relation to the shell-resolution fix and should be dropped from this PR.

**2. Missing test for `shell: null` from `os.userInfo()`**

`normalizeShell(null)` correctly returns `null`, and the code handles it, but there is no explicit test covering the case where `userInfo()` returns `shell: null` (as opposed to `shell: ''` which is tested). The `os.userInfo()` typings allow `shell` to be `null` on some platforms, so a test like the empty-string case but with `null` would close that gap.

No correctness bugs found. The core logic (`os.userInfo().shell` → `SHELL` env → `/bin/sh`) is sound, edge cases (throws, empty, non-executable) are handled and tested, and the DI-based testability design is clean.

johannesjo:
Thank you very much! <3

johannesjo (commented):
The `resolveUserShell()` implementation is correct and the tests are comprehensive. One minor scope issue: the diff includes `.prettierignore` entries that appear unrelated to the shell-resolution fix — please check if these were accidentally included from another branch and clean up the diff. Once that's resolved this is ready to merge. Details in the review comment.

ASRagab:
removed entries...thanks

johannesjo:
Thank you very much! <3

PR #61, opened 2026-04-07
fix(git): validate refs before worktree creation and improve main branch detection

FourWindff:
## Problem

1. `detectMainBranch()` only checks remote-tracking refs (`refs/remotes/origin/*`) when looking for `main`/`master`. In repositories without a remote (or before the first fetch), this causes the detection to fail even though a local `main` or `master` branch exists.

2. `createWorktree()` does not validate that the start-point ref exists before running `git worktree add`. When the ref is missing (e.g. in an empty repo or with a typo), Git produces a cryptic error message that is hard for users to diagnose.

## Solution

- Add a `localBranchExists()` helper that checks `refs/heads/<branch>`.
- Update `detectMainBranch()` to fall back to local branch refs when no remote-tracking ref is found.
- Add start-point ref validation in `createWorktree()` before attempting `git worktree add`, with clear error messages for:
  - Empty repositories (no commits yet)
  - Non-existent branch references

### DIFF
--- a/electron/ipc/git.ts
+++ b/electron/ipc/git.ts
@@ -147,6 +147,18 @@ async function remoteTrackingRefExists(repoRoot: string, branch: string): Promis
   }
 }
 
+/** Check whether a local branch ref exists. */
+async function localBranchExists(repoRoot: string, branch: string): Promise<boolean> {
+  try {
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much

---

PR #62, opened 2026-04-07
fix(electron): inherit full login shell environment instead of PATH only

FourWindff:
# Description

When launched from a `.desktop` file or AppImage, the app previously only inherited `PATH` from the user's login shell via `fixPath()`. All other environment variables were lost, which caused two issues:

1. **Missing environment variables in PTYs** — Environment variables like `GOPATH`, `JAVA_HOME`, `SSH_AUTH_SOCK`, and `NVM_DIR` were not available in spawned PTYs, causing tools that depend on them to fail.
2. **Claude unusable with third-party API keys** — Users who access Claude through a third-party API provider need custom environment variables (e.g. `ANTHROPIC_BASE_URL`) configured in their shell profile. Since only `PATH` was inherited, these variables were missing and Claude could not connect at all.

This PR replaces `fixPath()` with `fixEnv()`, which runs `env -0` inside an interactive login shell to capture **all** environment variables (null-byte separated), then applies them to `process.env`. This ensures spawned PTYs have the same complete environment as the user's terminal.

## Issues Resolved

- AppImage / `.desktop` launcher only inherited `PATH` from the login shell; all other environment variables were lost in spawned PTYs.
- Claude was unusable when configured with third-party API keys, because custom environment variables (e.g. `ANTHROPIC_BASE_URL`) were not passed through to the PTY.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -25,25 +25,33 @@ const __dirname = path.dirname(__filename);
 // welcome messages). Login-only (-lc) would be quieter but would miss tools
 // that are only added to PATH in .bashrc/.zshrc (e.g. nvm). We accept the
 // side effects since the sentinel-based parsing discards all other output.
-function fixPath(): void {
+function fixEnv(): void {
   if (process.platform === 'win32') return;
   try {
     const loginShell = resolveUserShell();
-    const sentinel = '__PCODE_PATH__';
-    const result = execFileSync(loginShell, ['-ilc', `printf "${sentinel}%s${sentinel}" "$PATH"`], {
-      encoding: 'utf8',
-      timeout: 5000,
-    });
-    const match = result.match(new RegExp(`${sentinel}(.+?)${sentinel}`));
-    if (match?.[1]) {
-      process.env.PATH = match[1];
+    const sentinel = '__PCODE_ENV__';
+    const result = execFileSync(
+      loginShell,
+      ['-ilc', `printf '${sentinel}' && env -0 && printf '${sentinel}'`],
+      { encoding: 'utf8', timeout: 5000 },
+    );
+    const startIdx = result.indexOf(sentinel);
+    const endIdx = result.lastIndexOf(sentinel);
+    if (startIdx === -1 || endIdx === -1 || startIdx === endIdx) return;
+
+    const envBlock = result.slice(startIdx + sentinel.length, endIdx);
+    for (const entry of envBlock.split('\0')) {
+      if (!entry) continue;
+      const eqIdx = entry.indexOf('=');
+      if (eqIdx <= 0) continue;
+      process.env[entry.slice(0, eqIdx)] = entry.slice(eqIdx + 1);
     }
   } catch (err) {
-    console.warn('[fixPath] Failed to resolve login shell PATH:', err);
+    console.warn('[fixEnv] Failed to resolve login shell environment:', err);
   }
 }
 
-fixPath();
+fixEnv();
 
 // Verify that preload.cjs ALLOWED_CHANNELS stays in sync with the IPC enum.
 // Logs a warning in dev if they drift — catches mismatches before they hit users.
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -32,7 +32,10 @@ function fixEnv(): void {
     const sentinel = '__PCODE_ENV__';
     const result = execFileSync(
       loginShell,
-      ['-ilc', `printf '${sentinel}' && env -0 && printf '${sentinel}'`],
+      [
+        '-ilc',
+        `printf '${sentinel}' && perl -e 'print "$_=$ENV{$_}\\0" for keys %ENV' && printf '${sentinel}'`,
+      ],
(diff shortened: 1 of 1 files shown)


johannesjo (changes requested):
## Code Review: fix(electron): inherit full login shell environment instead of PATH only

### Summary

This PR replaces `fixPath()` (which only captured `PATH` from the user's login shell) with `fixEnv()`, which runs `env -0` inside an interactive login shell to capture **all** environment variables. It also adds `localBranchExists()` and start-point ref validation to `git.ts` for safer worktree creation. The motivation is sound — missing env vars like `ANTHROPIC_BASE_URL`, `GOPATH`, `SSH_AUTH_SOCK` etc. are a real pain point when launching from `.desktop` files or AppImage.

---

### Blocking Issue: `env -0` breaks macOS

**This is the primary reason for requesting changes.**

`env -0` (null-byte-separated output) is a **GNU coreutils extension**. macOS ships BSD `env`, which does **not** support the `-0` flag. Running this on macOS will cause `env` to error out, which means `fixEnv()` will hit the `catch` block and silently fail — leaving `process.env` untouched. This is **worse** than the old `fixPath()` behavior, which worked correctly on both platforms.

Since the project explicitly targets **macOS and Linux** (per `CLAUDE.md`), this breaks one of the two supported platforms.

**Recommended fix — one of:**
1. **Gate `env -0` behind a Linux-only check** and keep the original `fixPath()` sentinel approach for macOS:
   ```ts
   if (process.platform === 'darwin') {
     fixPath(); // existing sentinel-based PATH-only approach
   } else {
     fixEnv(); // full env capture via env -0
   }
   ```
2. **Use a cross-platform alternative** that works on both macOS and Linux, e.g.:
   - `printenv` with newline parsing (but values can contain newlines, so fragile)
   - `env` without `-0` combined with a known-safe delimiter approach
   - Spawn `node -e "process.stdout.write(JSON.stringify(process.env))"` inside the login shell — this is fully cross-platform and handles all edge cases

---

### Security / Environment Pollution Concern

`fixEnv()` **blindly overwrites every key** in `process.env` with whatever the login shell emits. The old `fixPath()` was deliberately scoped to `PATH` only. The new approach means:

- Any variable set by shell startup scripts (including potentially harmful ones) propagates into the Electron process and every spawned PTY
- Electron-internal variables that were set *before* `fixEnv()` runs could be silently overwritten
- If a shell startup script `export`s something unexpected (e.g., `NODE_OPTIONS`, `ELECTRON_RUN_AS_NODE`, `LD_PRELOAD`), it could change Electron's behavior in surprising ways

**Suggestion:** Consider at minimum a **denylist** of keys that should never be overwritten:
```ts
const PROTECTED_KEYS = new Set(['ELECTRON_RUN_AS_NODE', 'NODE_OPTIONS', 'LD_PRELOAD', 'LD_LIBRARY_PATH']);
for (const entry of envBlock.split('\0')) {
  const eqIdx = entry.indexOf('=');
  if (eqIdx <= 0) continue;
  const key = entry.slice(0, eqIdx);
  if (PROTECTED_KEYS.has(key)) continue;
  process.env[key] = entry.slice(eqIdx + 1);
}
```

---

### Bundled Concerns: git.ts changes

The `git.ts` changes (local branch fallback in `detectMainBranch`, ref validation before `createWorktree`) are **unrelated** to the env fix. They are individually good changes (and were approved in the now-merged PR #61 which addresses the same concerns), but bundling them here makes the PR harder to review and creates potential merge conflicts.

**Suggestion:** Remove the `git.ts` changes from this PR, since PR #61 (now merged) already addresses the same issues. If this PR's git changes go beyond what #61 covers, please rebase onto main and resolve any overlap.

---

### Minor Issues

1. **Silent `catch` with no logging context:** The `catch` block logs `[fixEnv] Failed to resolve login shell environment:` which is good, but on macOS this will fire on every launch with a cryptic `env: illegal option -- 0` error. Users may not understand why. A platform-specific message or approach would help.

2. **`printf '${sentinel}'` uses single quotes inside double-quoted shell arg:** This is correct (the single quotes are passed to the shell), but it's subtle. A brief comment would help future maintainers.

---

### Overall

The core idea is valuable and addresses a real user pain point. However, the `env -0` incompatibility with macOS is a **blocker** — it would silently regress the env-resolution behavior on half the supported platforms. Please address the macOS compatibility issue, consider the env-pollution denylist, and ideally split out the unrelated git.ts changes.

---

johannesjo:
Thank you very much! <3

johannesjo (changes requested):
No new commits have been pushed since the CHANGES_REQUESTED review on 2026-04-12. The blocking issue remains: `env -0` is a GNU coreutils extension that does not exist on macOS BSD `env`, so `fixEnv()` will silently fail on every macOS launch — leaving `process.env` untouched. This is worse than the original `fixPath()` behavior which worked correctly on both platforms.

Please address the macOS compatibility issue (platform guard or cross-platform alternative) before requesting re-review.

johannesjo:
Thank you very much!

PR #63, opened 2026-04-07
feat(docker): support project-specific Dockerfiles

ASRagab:
## Summary
- detect and build project-specific Dockerfiles from `.parallel-code/Dockerfile`
- pass an explicit project build context into Docker builds instead of inferring from the Dockerfile directory
- track Docker source explicitly (`default | project | custom`) instead of inferring UI state from image-tag prefixes
- harden Docker image checks for unreadable custom Dockerfiles and clean up the new-task / Docker UI flow
- ignore repo-local `.worktrees/**` in Electron Vite dev mode so task creation does not trigger renderer reloads

## Validation
- npm run compile
- npm run lint
- npm test
- npm run typecheck

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -102,6 +102,7 @@ export enum IPC {
   CheckDockerAvailable = 'check_docker_available',
   CheckDockerImageExists = 'check_docker_image_exists',
   BuildDockerImage = 'build_docker_image',
+  ResolveProjectDockerfile = 'resolve_project_dockerfile',
 
   // System
   GetSystemFonts = 'get_system_fonts',
--- a/electron/ipc/pty.test.ts
+++ b/electron/ipc/pty.test.ts
@@ -77,6 +77,7 @@ vi.mock('node-pty', () => ({
 }));
 
 import {
+  buildDockerImage,
   DOCKER_CONTAINER_HOME,
   dockerImageExists,
   hashDockerfile,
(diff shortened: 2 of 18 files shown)


ASRagab:
Resolved the merge conflict with upstream/main and addressed all review comments:

1. **`fileNameFromPath` cross-platform split** — now splits on both `/` and `\` via `/[\\/]/`
2. **`resolveProjectDockerfile` file-type check** — uses `statSync().isFile()` instead of bare `existsSync` to reject directories and non-regular files
3. **`inferDockerSource` test coverage** — added cases for default image, project prefix, custom tag, and `undefined`
4. **Emoji accessibility** — added `aria-hidden="true"` to the folder emoji span

johannesjo:
Thank you very much! <3

johannesjo:
Reviewed the diff including the follow-up addressing all 4 Copilot comments (cross-platform path split, `statSync().isFile()`, `inferDockerSource` test coverage, `aria-hidden`). Those are all correctly resolved.

One new issue worth flagging:

**One-way Docker ratchet in `NewTaskDialog`**

```ts
createEffect(() => {
  if (skipPermissions() && store.dockerAvailable) {
    setDockerMode(true);
  }
});
```

This auto-enables Docker when "skip permissions" is checked, but has no inverse — unchecking skip-permissions leaves Docker mode on. The user would have to manually uncheck Docker mode. If this is intentional (i.e. skip-permissions always implies Docker), a comment would clarify the intent. If unintentional, the effect should also handle the reverse:

```ts
createEffect(() => {
  if (skipPermissions() && store.dockerAvailable) {
    setDockerMode(true);
  } else if (!skipPermissions()) {
    setDockerMode(false);
  }
});
```

(Or drop the effect entirely and rely on the existing "Enable Docker" suggestion UI that was already shown when skip-permissions was active.)

Everything else looks good — the `buildDockerImage` deduplication logic, IPC validation, `dockerSource` persistence migration, and cleanup handling in `createEffect` are all solid.

johannesjo (commented):
All four original Copilot review comments were correctly addressed. One UX issue before merge:

**One-way Docker ratchet** — in `NewTaskDialog`, a `createEffect` auto-enables Docker when 'skip permissions' is checked, but there is no inverse: unchecking 'skip permissions' leaves Docker enabled. This can silently enable Docker mode for users who toggle that checkbox and then change their mind. Consider adding the inverse side-effect, or making the coupling explicit in the UI. Details in the review comment.

The rest of the PR (Dockerfile detection, `DockerSource` migration, Vite watch fix) is well-implemented.

ASRagab:
@johannesjo last commment addressed

johannesjo:
Thank you very much! Can you maybe have a look at the conflicts?

johannesjo:
Thank you very much!

PR #94, opened 2026-05-02
[codex] Fix Codex arguments

brooksc:
## Summary

Updates Codex launch arguments now that `--full-auto` is deprecated.

- Replaces the default Codex skip-permissions args with `--dangerously-bypass-approvals-and-sandbox` so the UI's “Dangerously skip all confirms” mode keeps bypassing prompts.
- Updates the Arena Codex preset to use the same bypass flag.
- Migrates restored persisted Codex agent definitions that still contain `--full-auto`.
- Adds PTY spawn command debug logging so future launch args are visible in logs.
- Redacts Docker `-e KEY=value` env values from PTY spawn logs.
- Repairs the macOS `node-pty` `spawn-helper` executable bit after install so PTY spawning works after dependency installs.

## Root Cause

Codex CLI no longer supports the old `--full-auto` flag. Existing defaults and persisted task definitions could still pass that deprecated flag. The replacement needs to preserve the existing skip-permissions behavior, which maps to Codex's explicit bypass flag. Separately, the local macOS `node-pty` helper could be installed without execute permissions, causing `posix_spawnp failed` before Codex launched.

## Validation

- `npm run check`
- Verified a local Codex session launches and logs the expected bypass args when skip-permissions mode is enabled.

## Note

Codex supports safer sandbox modes such as `--sandbox workspace-write`, but this PR intentionally preserves the existing “Dangerously skip all confirms” behavior. A future UI improvement could let Codex users choose between safer workspace automation and full bypass mode.

### DIFF
--- a/electron/ipc/agents.ts
+++ b/electron/ipc/agents.ts
@@ -31,7 +31,7 @@ const DEFAULT_AGENTS: AgentDef[] = [
     command: 'codex',
     args: [],
     resume_args: ['resume', '--last'],
-    skip_permissions_args: ['--full-auto'],
+    skip_permissions_args: ['--dangerously-bypass-approvals-and-sandbox'],
     description: "OpenAI's Codex CLI agent",
   },
   {
--- a/electron/ipc/pty.test.ts
+++ b/electron/ipc/pty.test.ts
@@ -4,63 +4,66 @@ import path from 'path';
 import type { BrowserWindow } from 'electron';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
-const { mockExecFileSync, mockExecFile, mockChildProcessSpawn, mockPtySpawn } = vi.hoisted(() => {
-  const mockExecFileSync = vi.fn((command: string, args?: string[]) => {
-    if (command === 'which' && args?.[0] === 'nonexistent-binary-xyz') {
-      throw new Error('not found');
-    }
-    return '';
-  });
+const { mockExecFileSync, mockExecFile, mockChildProcessSpawn, mockPtySpawn, mockLogDebug } =
+  vi.hoisted(() => {
+    const mockExecFileSync = vi.fn((command: string, args?: string[]) => {
+      if (command === 'which' && args?.[0] === 'nonexistent-binary-xyz') {
+        throw new Error('not found');
+      }
+      return '';
+    });
 
-  const mockExecFile = vi.fn();
-  const mockChildProcessSpawn = vi.fn(() => ({
-    stdout: { on: vi.fn() },
-    stderr: { on: vi.fn() },
-    on: vi.fn(),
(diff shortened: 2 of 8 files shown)


johannesjo:
Thank you very much for this!

Code review notes:

1. `electron/ipc/agents.ts` and `src/arena/ConfigScreen.tsx` replace `--full-auto` with only `--sandbox danger-full-access`. That disables sandboxing, but it does not preserve the UI’s “Dangerously skip all confirms” behavior. Current Codex exposes `--dangerously-bypass-approvals-and-sandbox` for that mode, or this likely needs `--ask-for-approval never` alongside the sandbox flag. As written, the skip-permissions toggle can still leave Codex asking for approvals.

2. `electron/ipc/pty.ts` logs spawn args after Docker args have been expanded. In Docker mode those args include `-e KEY=value` entries from forwarded environment variables, so verbose PTY logging can expose API keys/tokens in logs. The new `redactedSpawnArgs()` only redacts `/bin/sh -c`, not Docker env flags. I’d mask all `-e` values, or avoid logging Docker args entirely.

brooksc:
Update pushed in `8a62848`.

This PR now has two layers of changes:

## 1st PR update: functional Codex launch fix

The first update fixed the actual Codex launch behavior after `--full-auto` stopped being supported:

- Replaced Codex `--full-auto` usage with `--dangerously-bypass-approvals-and-sandbox` so the existing “Dangerously skip all confirms” mode continues to skip approval prompts and sandboxing.
- Updated the default Codex agent skip-permissions args.
- Updated the Arena Codex preset to use the same bypass flag.
- Added migration for restored persisted Codex agent definitions that still contain `--full-auto`.
- Added PTY spawn command debug logging so future launch-argument problems are easier to diagnose.
- Redacted sensitive PTY spawn log content, including Docker env values and `/bin/sh -c` command strings.
- Added the macOS `node-pty` `spawn-helper` executable-bit repair after install.

## 2nd PR update: regression coverage

The second update is test-only coverage for the behavior fixed above:

- Added Docker spawn log redaction tests for `-e KEY=value`, `--env KEY=value`, and `--env=KEY=value` forms.
- Added shell spawn redaction coverage for `/bin/sh -c`, which protects Arena prompt-bearing commands from showing up in debug logs.
- Added persistence migration coverage for Codex `--full-auto` definitions.
- Added negative migration cases so non-Codex agents using `--full-auto` are unchanged, and already-current Codex definitions remain unchanged.

## Validation

- `npm test -- electron/ipc/pty.test.ts src/store/persistence.test.ts` — 36 tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- Commit and push hooks both ran `npm run check` successfully.

johannesjo:
Thank you very much for the adjustments! <3

PR #96, opened 2026-05-02
Share Docker agent auth across containers

brooksc:
## Summary

Docker-mode agents currently start with a fresh writable home each time, so users have to re-sign into Claude, Codex, Gemini, OpenCode, or Copilot for every new Docker container. That repeated authentication overhead makes Docker isolation less appealing, especially when creating multiple worktrees or trying short-lived tasks.

This PR adds an opt-in setting to share agent auth across Linux containers. When enabled, the app creates a user-owned host directory under `~/.parallel-code/agent-auth/<agent>` and bind-mounts it into that agent's config directory inside Docker, such as `/tmp/.codex` or `/tmp/.claude`. Signing in once inside a Docker container then persists for later Docker containers of the same agent type, including containers launched from different worktrees.

## Implementation

- Adds a persisted `shareDockerAgentAuth` setting and a checkbox in Settings.
- Forwards the setting through the terminal spawn IPC path with main-process validation.
- Mounts per-agent auth directories for known agent commands only:
  - `claude` -> `/tmp/.claude`
  - `codex` -> `/tmp/.codex`
  - `gemini` -> `/tmp/.gemini`
  - `opencode` -> `/tmp/.config/opencode`
  - `copilot` -> `/tmp/.config/github-copilot`
- Uses host bind directories instead of Docker named volumes so the container process, which runs as the host UID/GID, can write credentials on first login.
- Covers enabled, disabled, and unknown-agent behavior in the Docker PTY tests.

## Limitations

- This does not copy or seed credentials from existing host config directories like `~/.codex` or `~/.claude`; it persists credentials created inside Docker after the setting is enabled.
- Auth is shared per agent type, not per project or worktree.
- Only the known built-in agent command names get auth mounts. Custom or differently named wrappers are intentionally ignored unless added to the mapping.
- Existing running containers are unaffected; the mount is applied when a Docker agent is spawned.

## Validation

- `npm test -- electron/ipc/pty.test.ts`
- `npm run typecheck`
- `npm run lint -- --quiet`
- `npm run format:check`
- Commit and push hooks also ran `npm run check`.

### DIFF
--- a/electron/ipc/pty.test.ts
+++ b/electron/ipc/pty.test.ts
@@ -346,6 +346,21 @@ describe('spawnAgent docker mode', () => {
       expect(fs.existsSync(hostDir)).toBe(true);
     });
 
+    it('bind-mounts .claude.json file for claude so auth persists across containers', () => {
+      const home = makeTempHome([]);
+      vi.stubEnv('HOME', home);
+
+      spawnAgent(
+        createMockWindow(),
+        buildSpawnArgs({ command: 'claude', shareDockerAgentAuth: true }),
+      );
+
+      const volumeFlags = getFlagValues(getLastSpawnCall().args, '-v');
+      const expectedHostFile = `${home}/.parallel-code/agent-auth/claude/.claude.json`;
+      expect(volumeFlags).toContain(`${expectedHostFile}:${DOCKER_CONTAINER_HOME}/.claude.json`);
+      expect(fs.existsSync(expectedHostFile)).toBe(true);
+    });
+
     it('does not mount agent auth directory when shareDockerAgentAuth is disabled', () => {
       const home = makeTempHome([]);
       vi.stubEnv('HOME', home);
--- a/electron/ipc/pty.test.ts
+++ b/electron/ipc/pty.test.ts
@@ -126,6 +126,7 @@ function buildSpawnArgs(
     rows: 40,
     dockerMode: true,
     dockerImage: 'parallel-code-agent:test',
+    shareDockerAgentAuth: false,
     onOutput: { __CHANNEL_ID__: 'channel-1' },
     ...overrides,
   };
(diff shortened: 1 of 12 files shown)


johannesjo (commented):
## Review

Clean implementation — opt-in setting flows nicely from UI → store → IPC → backend, tests cover the important cases, and pre-creating the host directory to avoid root-ownership is a good detail.

### Issues

**1. `hostDir` is the same for all entries in an agent's config array** (`electron/ipc/pty.ts:616`)

Currently every agent has exactly one config dir, but the data structure is `Record<string, string[]>`. If an agent ever gets multiple entries, both container paths would be backed by the same host directory, causing cross-contamination.

Suggested fix — incorporate `relDir` into the host path:
```ts
const hostDir = path.join(home, '.parallel-code', 'agent-auth', agentCommand, relDir);
```

Or simplify the type to `Record<string, string>` since arrays aren't actually needed today.

**2. Command matching uses the raw `args.command` string** (`electron/ipc/pty.ts:567-573`)

If a user specifies a full path (e.g., `/usr/local/bin/claude`) or the command includes arguments, the `AGENT_CONFIG_DIRS` lookup silently fails. Consider matching against `path.basename(agentCommand)` or documenting that only bare command names are supported.

### Minor

- The `for...of` loop over a single-element array feels like dead code — either the type should be `string` or the host-dir path needs to account for multiple entries (see above).

Overall this is solid and well-tested. The first point is the only one I'd want addressed before merge.

johannesjo:

brooksc:
## Update

### Review issues addressed

All issues raised in the review have been addressed:

1. **`hostDir` ignores `relDir`** — fixed; host path now includes `relDir` so each config directory gets its own isolated location (e.g. `agent-auth/claude/.claude/`).
2. **Command matching uses raw `args.command`** — fixed; now uses `path.basename(agentCommand)` so full paths like `/usr/local/bin/claude` work correctly.
3. **`for...of` over single-element array / dead code** — resolved as a consequence of fix #1.
4. **`shareDockerAgentAuth` missing from autosave snapshot** — fixed; toggling the checkbox now immediately schedules a save.
5. **Host auth directory created with default permissions** — fixed; now uses `mode: 0o700`.
6. **`mkdirSync` throws and aborts Docker spawn** — fixed; failure now logs a warning and skips the mount instead of preventing the container from starting.

### Additional changes since the original PR

- **`.claude.json` file mount** — discovered during testing that Claude stores its main auth config (OAuth tokens) in `~/.claude.json` (a file at HOME level) in addition to `~/.claude/`. Without mounting this file, credentials written in the first container were lost when it exited and the next container prompted for login again. Added an `AGENT_CONFIG_FILES` mapping alongside `AGENT_CONFIG_DIRS` to handle file-level mounts, seeding the host file with `{}` on first use so Claude doesn't reject it as invalid JSON.
- **New task dialog message** — when "Share agent auth across Linux containers" is enabled, the Docker info banner now appends "Agent credentials are shared across containers."
- Rebased onto upstream `main`.

### Test results

✅ **Claude Code** — auth persists correctly across containers; no login prompt on second container.  
✅ **Codex** — auth persists correctly across containers.  
⚠️ **Gemini** — fails with `uv_os_get_passwd returned ENOENT` (exit code 41) before auth sharing is even reached. Root cause: Gemini CLI calls `os.userInfo()` to derive its keychain encryption key, and that libuv call reads `/etc/passwd` — which doesn't have an entry for the host UID (501) inside the container. This is unrelated to this PR and would occur with or without auth sharing enabled. Fix requires the Docker image to support arbitrary host UIDs, e.g. via [`fixuid`](https://github.com/boxboat/fixuid) or a dynamic `/etc/passwd` entry at container startup.  
🔲 **OpenCode** — not tested.  
🔲 **Copilot** — not tested.

johannesjo:
Thank you very much!

PR #97, opened 2026-05-02
feat(shortcuts): add Cmd+1-9 to jump to task by sidebar position

brooksc:
## Summary

- Adds `Cmd+1` through `Cmd+9` keyboard shortcuts to jump directly to the Nth task in the sidebar
- Shortcuts use the visual sidebar order (project-grouped, active before collapsed) so the number matches what the user sees
- Shortcuts are marked `global: true` so they fire even when a terminal has focus

## Implementation

- `src/store/navigation.ts` — new `jumpToTask(index)` function using `computeSidebarTaskOrder()`
- `src/lib/keybindings/defaults.ts` — 9 new bindings (`app.nav.jump-to-task-1` … `app.nav.jump-to-task-9`)
- `src/store/store.ts` — exports `jumpToTask` through the store barrel
- `src/App.tsx` — wires the 9 action strings to handlers

## Test plan

- [ ] Open app with 3+ tasks and press `Cmd+2` — second task in sidebar becomes active
- [ ] Press `Cmd+1` — first task becomes active
- [ ] Press `Cmd+9` with fewer than 9 tasks — no crash, no change
- [ ] Inside a terminal, press `Cmd+2` — still switches task (global shortcut)
- [ ] `npm test` — all 312 tests pass (5 new tests in `src/store/navigation.test.ts`)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,6 +24,7 @@ import {
   toggleSidebar,
   toggleArena,
   moveActiveTask,
+  jumpToTask,
   adjustGlobalScale,
   resetGlobalScale,
   startTaskStatusPolling,
@@ -452,6 +453,9 @@ function App() {
       'navigateColumn:right': () => navigateColumn('right'),
       'moveActiveTask:left': () => moveActiveTask('left'),
       'moveActiveTask:right': () => moveActiveTask('right'),
+      ...Object.fromEntries(
+        Array.from({ length: 9 }, (_, i) => [`jumpToTask:${i + 1}`, () => jumpToTask(i)]),
+      ),
       closeShell: () => {
         const taskId = store.activeTaskId;
         if (!taskId) return;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -50,7 +50,12 @@ import {
 } from './store/store';
 import { isGitHubUrl } from './lib/github-url';
 import type { PersistedWindowState } from './store/types';
-import { initShortcuts, registerFromRegistry, registerZoomShortcuts } from './lib/shortcuts';
+import {
+  initShortcuts,
+  registerFromRegistry,
+  registerJumpToTaskShortcuts,
+  registerZoomShortcuts,
+} from './lib/shortcuts';
 import { resolvedBindings, loadKeybindings, dismissMigrationBanner } from './store/keybindings';
 import { setupAutosave } from './store/autosave';
 import { isMac, mod } from './lib/platform';
@@ -520,6 +525,8 @@ function App() {
       resetZoom: () => resetGlobalScale(),
     });
 
+    const cleanupJumpToTaskShortcuts = registerJumpToTaskShortcuts((i) => jumpToTask(i));
+
     createEffect(() => {
       const cleanup = registerFromRegistry(resolvedBindings(), actionHandlers);
       onCleanup(cleanup);
@@ -539,6 +546,7 @@ function App() {
       unlistenResized?.();
       unlistenMoved?.();
       cleanupZoomShortcuts();
+      cleanupJumpToTaskShortcuts();
     });
   });
 
--- a/src/lib/keybindings/__tests__/defaults.test.ts
+++ b/src/lib/keybindings/__tests__/defaults.test.ts
@@ -24,6 +24,8 @@ const APP_LAYER_IDS = [
   'app.toggle-settings',
   'app.close-dialogs',
   'app.reset-zoom',
+  ...Array.from({ length: 9 }, (_, i) => `app.nav.jump-to-task-${i + 1}`),
+  ...Array.from({ length: 9 }, (_, i) => `app.nav.jump-to-task-${i + 1}-shift`),
 ];
 
 const TERMINAL_LAYER_IDS = [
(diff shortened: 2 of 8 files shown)


johannesjo:
Thank you very much for this!



# Code Review

Reviewed end-to-end (binding wiring, matchers, Linux behavior, tests). Two critical findings before this can merge — the headline one is that the shortcut never fires on any platform.

### Critical — shortcut never fires

`src/lib/keybindings/defaults.ts` uses ``key: `Digit${i + 1}` ``, but both matchers compare against `KeyboardEvent.key`, not `KeyboardEvent.code`:

- `src/lib/keybindings/match.ts:13` — `e.key.toLowerCase() \!== binding.key.toLowerCase()`
- `src/lib/shortcuts.ts:27` — same comparison

For the digit row, `e.key === "1"`; `"Digit1"` is the `code`. So `"1" === "digit1"` is always false and the binding never matches. The existing `app.reset-zoom` correctly uses `key: '0'` (`defaults.ts:260`) — same convention applies here.

**Fix:** ``key: `${i + 1}` ``.

### Critical — test does not exercise the binding

`src/store/navigation.test.ts` mocks `./core` and `./sidebar-order` and calls `jumpToTask()` directly. It proves the trivial `order[index]` lookup but never touches `matchesKeyEvent` / `matches` / the `App.tsx` handler map. That's why the `Digit1` bug slipped through.

**Suggestion:** add one integration test that dispatches a synthetic `KeyboardEvent("keydown", { key: "1", ctrlKey: true })` after `registerFromRegistry(DEFAULT_BINDINGS, handlers)` and asserts the handler is called.

### Important — Linux + non-US layouts

Once `key` is fixed to `"1"`…`"9"`, the binding still misses on layouts where the digit row requires Shift (AZERTY, several EU layouts). The existing `Cmd+0` reset-zoom binding documents this and registers a `shift: true` variant in `src/lib/shortcuts.ts:87-94`. Same treatment needed here — either (a) register shift variants for each digit, or (b) match on `e.code` for layout-independence (larger change; would ironically have justified the original `Digit1` value).

On Linux, `cmdOrCtrl` resolves to Ctrl (`match.ts:19-20`), so this intercepts `Ctrl+1`…`Ctrl+9` globally — including inside xterm.js (`TerminalView.tsx:237`). Follows the precedent of `Ctrl+0`, but the test plan only covers Cmd+1/2/9 and should mention Linux explicitly.

### Minor

- `src/store/sidebar-order.ts:40-49` includes collapsed tasks in the order — `Cmd+1` may activate a task hidden inside a collapsed project, which is surprising. Worth confirming intent and adding a test for the collapsed case.
- `src/App.tsx:456-458` — spreading nine generated entries into `actionHandlers` works but reads differently from surrounding static entries; nine explicit entries (or a single `jumpToTask` action that parses a numeric arg) would be more consistent.
- `defaults.ts:79-89` — nine near-identical "Jump to task N" descriptions will look noisy in any keybindings UI.

### Verdict

Needs changes. After the `key` fix + an integration test + the shift-variant story for non-US layouts, this is a small, low-risk PR.

brooksc:
## Changes in response to review

**Critical fix — shortcut never fired:** Changed `key: \`Digit${i+1}\`` to `key: \`${i+1}\`` in `defaults.ts`. Matchers compare against `KeyboardEvent.key` (e.g. `"1"`), not `KeyboardEvent.code` (e.g. `"Digit1"`), so the binding never matched.

**AZERTY / non-US layout support:** Added shift variants for each digit (e.g. `app.nav.jump-to-task-1-shift` with `modifiers: { cmdOrCtrl: true, shift: true }`), following the same pattern already used for the `Cmd+0` reset-zoom shortcut.

**Integration tests:** Added two tests to `shortcuts.test.ts` that go through the full `registerFromRegistry → initShortcuts → keydown` path: one confirming `key="1"` fires the handler, one confirming `key="Digit1"` does not (regression guard for the original bug).

Tested and working — Cmd+1 through Cmd+9 now correctly jump to tasks by sidebar position.

johannesjo:
Thank you very much!


Pushed `8738eba` directly to the branch — two changes, both UX-focused:

**1. `Cmd+N` now indexes `store.taskOrder`, not the project-grouped sidebar order.**

The previous `jumpToTask` walked `computeSidebarTaskOrder()`, which regroups tasks by project (per-project active, then per-project collapsed, then orphans). When projects are interleaved in `taskOrder`, that order diverges from the L→R tile layout in the main area and from the order `Cmd+Left/Right` cycles through. So pressing `Cmd+2` could activate a different task than the visually second tile, which is hard to predict and inconsistent with the rest of the navigation model.

Now all three navigation paths agree: `Cmd+1..9`, `Cmd+Left/Right`, and tile order are the same sequence. As a side effect, `taskOrder` excludes collapsed tasks, so `Cmd+N` can no longer activate a task hidden inside a collapsed project group (the Minor #1 from the earlier review).

**2. AZERTY shift variants moved out of the keybindings registry.**

The shift variants in `defaults.ts` made the Keyboard Shortcuts dialog show 18 jump-to-task rows (9 with `Cmd+N`, 9 duplicate "Jump to task N" rows with `Cmd+Shift+N`). They now live in a new `registerJumpToTaskShortcuts()` in `shortcuts.ts`, mirroring how the `Cmd+0` reset-zoom shift variant is registered imperatively rather than via the registry. The UI is back to 9 rows.

**Tests:** added one unit test pinning `jumpToTask` ignores `collapsedTaskOrder`, and one integration test for the AZERTY path (`key="1", shiftKey: true`). All 316 tests pass; lint/typecheck/prettier clean.

Not included: the sidebar `1`–`9` indicator polish (touches `Sidebar.tsx` rendering — more a polish change than a correctness fix, happy to do as a follow-up).

johannesjo:
Thank you! <3

PR #31, opened 2026-03-24
Feat: Coordinating Agent

cledoux95:
## Summary

- Add an MCP server (`electron/mcp/`) with tools (`create_task`, `send_prompt`, `wait_for_idle`, `get_task_diff`, `merge_task`, etc.) enabling a Claude Code agent to programmatically orchestrate tasks in Parallel Code
- Add main-process orchestrator, REST API endpoints, and MCP client infrastructure
- Add UI support: coordinator mode toggle in new task dialog, sub-task status strip, "via Coordinator" sidebar labels, and consolidated `TaskPanel` component
- Refactor: merge `TaskAITerminal`, `TaskBranchInfoBar`, `TaskClosingOverlay`, `TaskNotesPanel`, `TaskShellSection`, and `TaskTitleBar` into a single `TaskPanel`
- Remove Docker isolation mode, `SegmentedButtons`, system fonts IPC, and theme utilities
- Simplify `NewTaskDialog`, `SettingsDialog`, `ScrollingDiffView`, and multiple store modules

## Test plan

- [ ] Verify MCP server starts and responds to tool calls
- [ ] Create a task via the coordinating agent and confirm it appears in the UI
- [ ] Test `send_prompt`, `wait_for_idle`, `get_task_diff`, and `merge_task` end-to-end
- [ ] Confirm sub-task strip renders correctly with multiple coordinated tasks
- [ ] Verify standard (non-coordinated) task creation still works
- [ ] Confirm removed components (Docker mode, system fonts, segmented buttons) have no lingering references

PR #98, opened 2026-05-04
feat(mcp): pilot/co-pilot control handoff for coordinated tasks

brooksc:
## Summary

Adds explicit human/orchestrator control handoff for sub-tasks created by a coordinating agent, modelled on a pilot/co-pilot system where control is always unambiguously held by one party.

This is a standalone feature built on top of the coordinating agent concept introduced in PR #31.

### How it works

Each sub-task created by a coordinator has a runtime field `controlledBy: 'orchestrator' | 'human'`. The task panel shows a persistent banner so the user always knows who has the stick:

- **Orchestrator driving** — subtle bar with a "Take Control" button
- **Human in control** — amber warning banner: *"You have control — orchestrator is paused"* with a "Return to Orchestrator" button

When the human holds control, the orchestrator's MCP tools are blocked for that task:
- `send_prompt` throws immediately so the coordinator agent knows it cannot proceed and must wait or work on other tasks
- `wait_for_idle` resolves immediately (rather than hanging) so the coordinator isn't blocked

Control returns to the orchestrator **only via explicit user action** — never automatically. Returning control also fires any queued `waitForIdle` resolvers so the coordinator can resume cleanly.

### Changes

- `src/store/types.ts` — add `controlledBy?: 'orchestrator' | 'human'` to `Task` (runtime only, not persisted)
- `electron/ipc/channels.ts` — add `MCP_ControlChanged` channel
- `src/store/tasks.ts` + `src/store/store.ts` — add `setTaskControl(taskId, who)` action
- `electron/mcp/orchestrator.ts` — enforce control in `sendPrompt` and `waitForIdle`
- `electron/ipc/register.ts` — handle `MCP_ControlChanged` IPC
- `src/components/TaskPanel.tsx` — control banner UI with take/return buttons

## Test plan

- [ ] Create a coordinator task; verify sub-tasks show "Orchestrator driving" bar
- [ ] Click "Take Control" on a sub-task; verify banner changes to amber warning
- [ ] While in human control, verify coordinator's `send_prompt` MCP call returns an error
- [ ] Click "Return to Orchestrator"; verify banner reverts and coordinator can send prompts again
- [ ] Verify non-coordinated tasks show no banner

### DIFF
--- a/COORDINATING-AGENT.md
+++ b/COORDINATING-AGENT.md
@@ -1,0 +1,128 @@
+# Coordinating Agent — Implementation Notes
+
+## Overview
+
+A Claude Code agent running inside a task can now programmatically create and coordinate other tasks in Parallel Code via MCP tools. The coordinating agent can spawn sub-tasks, send prompts, monitor completion, read diffs, and merge results — all without manual human orchestration.
+
+## Architecture
+
+```
+Claude Code (coordinating task PTY)
+    | MCP stdio (JSON-RPC)
+    v
+MCP Server Process (electron/mcp/server.ts)
+    | HTTP + Bearer token
+    v
+Electron Remote Server (electron/remote/server.ts)
+    | Direct function calls
+    v
+Orchestrator (electron/mcp/orchestrator.ts)
+    | Uses existing backend primitives
+    v
+PTY / Git / Worktree functions
+```
+
+## MCP Tools Available to Coordinating Agents
+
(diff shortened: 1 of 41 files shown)


brooksc:
ignore this PR -- I'm trying to create an update to PR 31... but without all the other PRs merged in (which I was using for my own purposes).  Another PR incoming, 3...2..1.. :)

PR #105, opened 2026-05-07
feat(inline-input): add cancel button and global Esc to dismiss

FourWindff:
# Summary                                                                                                                                                       
                                                                                                                                                                
The InlineInput.tsx component shown after selecting code in the diff viewer could not be cancelled once the input lost focus. This PR makes it always  dismissable.
                                                                                                                                                                
  - Add a global Escape keydown listener (on document, capture phase) so users can dismiss the popup at any time                                                
  - Add an explicit cancel button (×) next to the submit button for mouse users
  - Clean up the keydown listener on unmount


johannesjo:
Thank you very much! <3

PR #116, opened 2026-05-16
feat(themes): Themes settings tab with Light/Dark/System appearance mode

brooksc:
## Summary

This PR adds a **Themes** tab to the Settings dialog that lets users choose their appearance mode (Light / Dark / System) and pick independent theme presets for each mode. It also includes a contrast fix for three built-in themes and a reliability fix for the close-dialog fallback timer.

---

### Themes settings tab (`SettingsDialog.tsx`, `look.ts`, `ui.ts`, `store/`)

- New **Themes** tab in Settings with three appearance-mode options: **Light**, **Dark**, and **System** (follows the OS preference via `window.matchMedia('prefers-color-scheme: dark')`).
- Independent theme preset selectors for the dark slot and the light slot — switching modes instantly applies the preset last chosen for that mode.
- State is persisted per slot (`appearanceMode`, `darkThemePreset`, `lightThemePreset`, `darkThemeCustomId`, `lightThemeCustomId`).
- Backward-compatible migration: existing users who had the old single `themePreset` field saved will have it migrated into the correct dark/light slot automatically on first load.

### WCAG AA contrast fix (`styles.css`)

Three built-in themes had accent colours that failed WCAG AA (4.5:1) against their panel backgrounds:

| Theme | Old colour | New colour | Old ratio | New ratio |
|---|---|---|---|---|
| classic | `#4c6fff` | `#4267ff` | 4.17:1 | 4.55:1 |
| islands-dark | `#548af7` | `#286cf5` | 3.30:1 | 4.61:1 |
| islands-light | `#3574f0` | `#2c6def` | 4.27:1 | 4.60:1 |

### Close-dialog fallback timer fix (`window.ts`)

When the "Cancel option" close dialog feature was added, `WindowCloseHandling` (which cancels the backend's 5 s force-destroy fallback) was being sent immediately on every close request — before the async pre-work (`captureWindowState`, `saveState`, `CountRunningAgents`) had completed. A hang in any of those left the window permanently stuck because the safety net was already gone.

Fixed by moving the ack into the `preventDefault()` callback so the timer stays armed through the pre-work and is only cleared once the renderer has decided to show the interactive dialog.

### Tests

- `isLookPreset` type guard: valid preset IDs, unknown strings, non-string values.
- Theme persistence migration: defaults, all three appearance modes, invalid-mode fallback, valid/invalid preset restore, custom ID handling, and all three backward-compat paths for the legacy `themePreset` field (13 new tests, 475 total passing).

---

## How to test

1. Open Settings → Themes tab.
2. Switch between Light / Dark / System — the app should immediately apply the corresponding preset.
3. Pick a different preset in Dark mode, switch to Light, pick one there, switch back — each mode remembers its own choice.
4. Quit and relaunch — selections should be restored.
5. (Migration) Manually set `"themePreset": "classic"` in the saved state file and relaunch without `appearanceMode` — should restore dark mode with Classic.


brooksc:
image

Shows the Settings on the General Tab 

image

When user selects Theme, if they select Light (always light them) it shows the Light themes

image

When user selects Theme, if they select Dark (always dark them) it shows the Dark themes

image

When the user selects Theme, then System they can select both a light and dark theme.  Which one is currently enabled is based on their system controls - e.g. on Mac it will transition from Light to Dark in the evening and back to Light in the morning.

johannesjo:

Issue #3, opened 2026-02-21
[Feature Request] Support for non-interactive mode (e.g., --dangerously-skip-permissions for Claude)


Iktahana:
### Foreword

Thank you for taking the time to fill this feature request fully. Without it we may not be able to address it, and the issue may be closed without resolution.

## Problem Statement

When running tasks with `claude-code`, the agent frequently pauses execution to ask: **"Do you want to proceed? (y/N)"**. 

This is very frustrating because it breaks the "parallel" and automated nature of the app. I often start a task and step away (e.g., to get a coffee), but when I return, I find the agent has been stuck for 10 minutes waiting for me to manually type "y". This issue is even more frequent when using a proxy or when the agent needs to perform multiple file operations.

##  Possible Solution

I would like the ability to run agents in a non-interactive/unattended mode. Specifically, for Claude, I would like to be able to pass the `--dangerously-skip-permissions` flag.

Possible implementations:
1. **Global Toggle:** A setting to "Always skip permissions (dangerous)" in the app settings.
2. **Custom CLI Flags:** A field in the "New Task" dialog that allows me to append custom arguments to the agent command.

## Describe alternatives you've considered

* **Manual Monitoring:** Staying at the desk just to press "Enter," which defeats the purpose of an automated coding assistant.
* **Piping "yes":** I tried to think of ways to pipe "yes" to the terminal, but that is unreliable and often breaks the terminal UI (xterm.js) within the app.

## Additional context

The `claude` CLI introduced the `--dangerously-skip-permissions` flag specifically to solve this "stuck on prompt" issue for automation. Since Parallel Code is all about managing multiple agents efficiently, having a way to bypass these interactive gates would significantly improve the user experience for long-running background tasks.

**Thank you so much for this amazing project! It’s incredibly useful and has quickly become a core part of my workflow—I absolutely love it!**

johannesjo:
Thank you very much. I will add this as a checkbox to the add task dialog.

PR #127, opened 2026-05-20
Fix status dot tooltip behavior

brooksc:
## Problem

Status dots appear in multiple places, including normal sidebar rows, collapsed sidebar rows, and the task title bar. Before this PR, those dots did not have tooltips, so users had to infer what each color meant.

Because the dot is reused in several UI locations, the tooltip needs to live on the shared `StatusDot` component instead of being implemented separately at each call site. That also keeps the tooltip text consistent with attention-state overrides such as active, waiting for input, or error.

The branch also intended review-ready steps to take precedence over active/busy state, but `getTaskDotStatus()` checked activity before the latest step status. A task with an active agent and a latest `awaiting_review` step could still show as busy instead of ready for review.

## Solution

Add tooltip ownership to the reusable `StatusDot` component so every dot renders the same explanation from the same status and attention inputs. Update task dot status priority so the latest `awaiting_review` step returns `review` before active task activity can return `busy`.

## Example

Status dot tooltip example

## Changes Made

1. Added `getDotTooltip()` to map dot status and attention state to user-facing tooltip text.
2. Added the tooltip `title` directly to `StatusDot` using `getDotTooltip(status, attention)`.
3. Reordered `getTaskDotStatus()` so latest-step `awaiting_review` wins over busy activity.
4. Added a regression test for review-before-busy status precedence.
5. Added tests for tooltip text mapping and the rendered `StatusDot` title attribute.
6. Updated Vitest config to run the existing Solid transform in SSR mode under the node test environment, enabling Solid component rendering tests without adding jsdom.
7. Added the tooltip screenshot used above as a PR example asset.

## Verification

- `npm test -- src/store/taskStatus.test.ts src/components/StatusDot.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `git diff --check`


johannesjo:
Thank you very much!



## Review

Multi-agent review (correctness, security, architecture, alternatives, performance, simplicity). Verified against tree at `e97cce3`.

### Blocking

**1. Stated "review beats active" intent does not render at `TaskTitleBar`.**
`src/components/TaskTitleBar.tsx:85-89` passes both `status={getTaskDotStatus(...)}` and `attention={getTaskAttentionState(...)}`. With the reorder, a task with active agent + latest `awaiting_review` step now returns `status='review'` AND `attention='active'`. Inside `StatusDot`:
- `getDotColor` (`StatusDot.tsx:7`) hits `attention === 'active'` first → paints `theme.accent` (active blue), not review purple.
- New `getDotTooltip` (`StatusDot.tsx:28`) hits the same branch first → returns `"Active — agent is working"`, not `"Ready for review"`.
- `isPulsing` (`StatusDot.tsx:33`) still pulses on `attention === 'active'`.

The reorder in `getTaskDotStatus()` only surfaces at sidebar rows where `offscreenAttention.attention()` is `undefined` for visible tasks. The new test (`taskStatus.test.ts:684-695`) asserts the store return value, not rendered output, so it passes while the title-bar dot still looks/behaves active.

Pick one:
- Drop `attention={getTaskAttentionState(...)}` at `TaskTitleBar.tsx:88` — let the dot reflect dot-status only.
- Move review precedence into `StatusDot` so `status === 'review'` overrides `attention === 'active'` in `getDotColor` + `getDotTooltip` + `isPulsing` symmetrically.

### Important

**2. `getDotTooltip` is not exhaustive at the type level.** `StatusDot.tsx:50-55` — signature returns `string` but the inline lookup is `string | undefined` to TS. A future `TaskDotStatus` variant silently yields `title={undefined}` with no compile error. Fix:

```ts
const STATUS_TOOLTIP = {
  busy: 'Busy — agent recently active',
  waiting: 'Waiting — no changes yet',
  ready: 'Ready to merge',
  review: 'Ready for review',
} as const satisfies Record<TaskDotStatus, string>;
```

Hoisted to module scope, also avoids per-call object allocation. Same gap exists in pre-existing `getDotColor` — out of scope but worth a follow-up.

**3. Vitest config switch is global for one test file.** `vitest.config.ts:4-7` applies `solidPlugin({ ssr: true })` to every `.test.ts` under `src/**` and `electron/**`. Four existing tests already `vi.mock('solid-js', ...)` (`taskStatus.test.ts:32`, `focus.test.ts:53`, `ui.test.ts:35`, `core.test.ts:3`). Plugin transforms `solid-js` imports; interaction with `vi.mock` ordering is non-obvious. `environment: 'node'` is the vitest default — that line is a no-op. Please confirm `npm test` (full suite, not just the two files in the PR body) still passes locally. Alternatives:
- Scope the plugin to component tests via a separate vitest project.
- Or drop the render test entirely (see #5) and revert this config.

**4. Triplicated `(status, attention) → derived value` logic.** `StatusDot.tsx` now has `getDotColor` (lines 6-14), `getDotTooltip` (45-55), `isPulsing` (33), all encoding the same enum precedence. Drift is already visible: `isPulsing` pulses on `status === 'busy'` while `getDotColor` paints `busy` as `theme.fgMuted` with no shadow → muted pulsing dot. One source of truth would help:

```ts
const STATUS_META = {
  busy:    { color: theme.fgMuted, tooltip: 'Busy — agent recently active', pulse: true },
  waiting: { color: '#e5a800',     tooltip: 'Waiting — no changes yet',     pulse: false },
  ready:   { color: theme.success, tooltip: 'Ready to merge',               pulse: false },
  review:  { color: '#c084fc',     tooltip: 'Ready for review',             pulse: false },
} as const satisfies Record<TaskDotStatus, DotMeta>;
```

**5. Test bypasses Solid prop reactivity.** `StatusDot.test.ts:22` calls `StatusDot({ status: 'review' })` as a plain function — works for static SSR but breaks if the component ever uses `splitProps`/`mergeProps`/`Show`/destructured defaults. Two clean options:
- Rename to `.test.tsx`, use JSX: `renderToString(() => <StatusDot status="review" />)`.
- Keep `getDotTooltip` private (drop `export`), delete the render test, assert tooltip text via direct unit calls only. This also resolves #3 by reverting `vitest.config.ts`.

The three direct `getDotTooltip` unit tests (lines 6-19) plus the render test (lines 21-27) overlap; one of them is enough.

### Minor

**6. Accessibility: `title` is mouse-only.** `StatusDot.tsx:50`. Non-interactive `<span>` with `title=` is widely ignored by screen readers and absent on touch. Strict improvement at zero LOC cost:

```tsx
<span
  role="img"
  aria-label={getDotTooltip(props.status, props.attention)}
  title={getDotTooltip(props.status, props.attention)}
  ...
/>
```

**7. Tooltip strings belong with the enum.** `TaskDotStatus` / `TaskAttentionState` live in `src/store/taskStatus.ts:108-109` as domain vocabulary. Their labels are properties of the state, not of one visualization. Exporting `taskStatusLabels` / `taskAttentionLabels` from `taskStatus.ts` keeps single-source-of-truth and makes future i18n a one-file change.

**8. Screenshot PNG.** Consistent with the existing `screens/` precedent (54MB / 21 binary assets), so not blocking. For one-off PR-body illustrations going forward, dragging into the GitHub PR body uploads to `user-attachments.githubusercontent.com` for free with no permanent clone weight.

### What's good

- Reorder in `getTaskDotStatus` (`taskStatus.ts:839-855`) is in the right layer and the store-level regression test (`taskStatus.test.ts:684-695`) is the load-bearing addition.
- Step-content null/empty guards preserved correctly.
- Centralizing tooltip ownership on the shared component is the right architectural call — just needs the precedence inconsistency at the title bar resolved (#1).
- No security concerns: inputs are closed string-literal unions, SolidJS sets `title` via `setAttribute` (no HTML sink).

### Verdict

Needs changes. Blocking item is #1 — PR description claims a fix the rendered title bar does not show. Items 2-5 ideally bundled into the same revision; 6-8 optional / follow-up.

johannesjo:
Thank you very much!

Issue #91, opened 2026-04-29
Inlcude check for updates and update without reinstalling new version


dserik20:
Hi,  can you please add check for updates and update without reinstalling new version.

VitorRibeiroCustodio:
@johannesjo are you currently working on this? If not, I'd like to volunteer to take it on — happy to wire up `electron-updater` with the GitHub releases provider for auto-update support.

johannesjo:
No, currently I am not working on this! Would be really great to have this! Thank you @VitorRibeiroCustodio !

PR #133, opened 2026-05-22
feat: in-app auto-update

VitorRibeiroCustodio:
## What

In-app auto-update — users move to a newer version without deleting and reinstalling the app (#91).

- Silent update check ~10s after launch; manual **Check for updates** in Settings → Diagnostics → Updates.
- A **sidebar header button** appears whenever an update is available / downloading / downloaded — the single apply path: one click downloads, then becomes restart & install. Hidden when up to date, unsupported, or the last check failed.
- The Settings Updates section shows the current version, a manual check, and the last-check outcome (up to date / available / error). It no longer carries its own apply buttons — applying happens via the sidebar button, so there is one apply path.
- Auto-update is a packaged-build capability. Dev runs and the Linux `deb` target (no in-place update channel) report "unavailable" instead of failing.
- Once downloaded, **Restart & install** relaunches onto the new version; a deferred install also applies automatically on next quit.

## How

- New backend module `electron/ipc/updater.ts` wraps `electron-updater`'s `autoUpdater` behind new IPC channels: `check_for_updates`, `download_update`, `quit_and_install_update`, `get_update_status`, and a pushed `update_status_changed` event.
- Renderer store slice `src/store/updates.ts` (status signal + subscription + check/download/install helpers); new `UpdateButton` component in the `Sidebar` header; Updates section in `SettingsDialog`.
- `electron-builder` config: GitHub `publish` target + a macOS `zip` artifact (electron-updater requires `zip` to apply macOS updates).
- No persisted-state or schema change — update status is transient, re-derived from the main process each launch.

## Testing

Verified locally against a generic update server: a packaged build detected a newer published version, downloaded it with progress, and relaunched onto the new version via Restart & install.

## Screenshots

Screenshot 2026-05-22 at 06 15 31

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ dist-electron
 dist-remote
 release
 coverage
+# Local auto-update test server payload (build artifacts, not source).
+update-test/
 # Runtime data dir written by the app (MCP coordinator state, etc.)
 .parallel-code/
 .worktrees
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -141,6 +141,13 @@ export enum IPC {
   // Logging
   LogFromRenderer = 'log_from_renderer',
 
+  // Auto-update
+  CheckForUpdates = 'check_for_updates',
+  DownloadUpdate = 'download_update',
+  QuitAndInstallUpdate = 'quit_and_install_update',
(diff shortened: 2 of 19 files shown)


johannesjo:
Thank you very much! <3

PR #135, opened 2026-05-24
Fix sidebar selection when jumping with Cmd-number

brooksc:
## Problem
When a task is selected with the mouse, the sidebar keeps focus and renders the selection outline from `sidebarFocusedTaskId`. Using Cmd-1 through Cmd-9 changed the active task, but left `sidebarFocusedTaskId` pointing at the last mouse-selected row, so the outline stayed on the wrong task.

## Fix
Update `jumpToTask()` so Cmd-number jumps also synchronize the sidebar focused task when the sidebar currently owns focus. The change clears project focus and moves the sidebar outline to the jumped task.

## Verification
- `npm run check`
- `npm test`
- Confirmed the PR branch contains one commit on top of latest `johannesjo/main`, touching only `src/store/navigation.ts` and `src/store/navigation.test.ts`.

### DIFF
--- a/src/store/navigation.test.ts
+++ b/src/store/navigation.test.ts
@@ -9,6 +9,9 @@ type MockStore = {
   collapsedTaskOrder: string[];
   projects: Array<{ id: string }>;
   focusedPanel: Record<string, string>;
+  sidebarFocused: boolean;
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #136, opened 2026-05-24
fix(sidebar): restore status tooltip hit area

brooksc:
## Problem
Status tooltips were moved from the sidebar task row onto only the 6-8px status dot. The tooltip still existed in the DOM, but users had to hover almost exactly over the tiny dot, so row hover no longer exposed the status explanation shown in the existing screenshot.

## Fix
Restore row-level status tooltip titles for sidebar task rows and coordinator rows, while keeping dot-level titles for compact usages. Sidebar status dots now receive task attention state so tooltip text matches states like waiting for input, active, or error.

## Verification
- `git diff --check`
- `/Users/brooksc/.codex/skills/parallel-code-pr/scripts/run-pr-verification.sh` (`npm run check` and `npm test`)
- Push hook reran `npm run check` and `npm test`
  - `src/components/Sidebar.tsx`
  - `src/components/StatusDot.test.ts`

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -39,7 +39,7 @@ import { ImportWorktreesDialog } from './ImportWorktreesDialog';
 import { SidebarFooter } from './SidebarFooter';
 import { IconButton } from './IconButton';
 import { UpdateButton } from './UpdateButton';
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #137, opened 2026-05-24
Fix coordinator control handoff

brooksc:
## Problem
Coordinator control handoff could stall or diverge across UI/backend state. Releasing human control only notified the coordinator after a blocked send_prompt, not after wait_for_idle was interrupted. The renderer also persisted optimistic control state before backend acknowledgement, so frontend and backend ownership could drift.

## Fix
Track idle waits interrupted by human control and stage the same resume notification when control returns to the coordinator. Make backend control changes fail loudly for missing coordinator state or unknown tasks. Persist renderer control state only after backend acknowledgement, with rollback on failure, and make auto-trust release wait for backend acknowledgement before updating UI state.

## Verification
- `git diff --check`
- `npm run check`
- `npm test`
  - `electron/ipc/register.ts`
  - `electron/mcp/coordinator.test.ts`
  - `electron/mcp/coordinator.ts`
  - `src/store/taskStatus.ts`
  - `src/store/tasks.test.ts`
  - `src/store/tasks.ts`

### DIFF
--- a/electron/ipc/register.ts
+++ b/electron/ipc/register.ts
@@ -1174,7 +1174,8 @@ export function registerAllHandlers(win: BrowserWindow): void {
         if (args.controlledBy !== 'coordinator' && args.controlledBy !== 'human') {
           throw new Error(`Invalid controlledBy: ${String(args.controlledBy)}`);
         }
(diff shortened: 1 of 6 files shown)


johannesjo:
Thank you very much! <3

PR #138, opened 2026-05-24
Fix terminal PTY interruptions during remounts

brooksc:
## Problem
Task AI terminals currently terminate their backend PTY when the TerminalView component cleans up. In dev/HMR or other renderer remounts, that kills an active Codex session, then the renderer respawns it with resume args. Codex sees that as an interrupted conversation and waits for manual continuation.

## Fix
Task AI terminals now preserve their PTY across component cleanup and reattach to the existing process on remount. Intentional replacement remains explicit: restart and agent switch set attachExisting=false, so those actions still replace the running process. Standalone close/cleanup paths continue to own termination.

## Verification
- git diff --check
- npm run check
- npm test
- Push hook reran npm run check and npm test
- Confirmed this branch is rebased on latest johannesjo/main
  - electron/ipc/pty.test.ts
  - electron/mcp/coordinator.ts
  - src/components/TaskAITerminal.tsx
  - src/components/TerminalView.tsx
  - src/store/agents.test.ts
  - src/store/agents.ts

### DIFF
--- a/electron/ipc/pty.test.ts
+++ b/electron/ipc/pty.test.ts
@@ -755,6 +755,30 @@ describe('spawnAgent session reattach', () => {
     ).not.toThrow();
     expect(mockPtySpawn).toHaveBeenCalledTimes(1);
   });
+
+  it('replaces an existing same-id PTY when attachExisting is explicitly false', () => {
+    const win = createMockWindow();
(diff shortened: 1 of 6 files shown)


johannesjo:
Thank you!

PR #140, opened 2026-05-24
Fix input handoff notification copy

brooksc:
## Problem
The coordinator handoff notification was hard-coded to say "Claude needs input" even when the running agent is not Claude.

## Fix
Pass the task name into `PromptInput` and use it in the handoff notification copy so the message is generic to the task being controlled.

## Verification
- `git diff --check`
- `npm run check`
- `npm test`
  - `src/components/PromptInput.tsx`
  - `src/components/TaskPanel.tsx`

Supersedes #139, which GitHub closed when the fork branch was renamed.

### DIFF
--- a/src/components/PromptInput.tsx
+++ b/src/components/PromptInput.tsx
@@ -39,6 +39,7 @@ export interface PromptInputHandle {
 
 interface PromptInputProps {
   taskId: string;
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you! <3

PR #141, opened 2026-05-25
Fix new-task panel UI interactions

FourWindff:
## Summary

Small fixes for the new-task dialog panel:

- Prevent accidental text selection in the dialog
- Prevent the base-branch label from stretching and triggering the dropdown when clicked outside the intended area


johannesjo:
Thank you very much! <3

PR #142, opened 2026-05-25
fix(coordinator): default coordinator tasks to human control

FourWindff:
## Summary
- initialize newly created coordinator tasks with human control instead of coordinator control
- when restoring persisted tasks without a control owner, default coordinator tasks to human control and coordinated child tasks to coordinator control
- add persistence tests for the missing-controlledBy restore defaults

### DIFF
--- a/src/store/persistence.test.ts
+++ b/src/store/persistence.test.ts
@@ -369,6 +369,48 @@ describe('coordinator control hint persistence', () => {
   });
 });
 
+describe('coordinator control defaults', () => {
+  it('restores coordinator tasks to human control when persisted state omits controlledBy', async () => {
+    mockInvoke.mockResolvedValueOnce(
+      basePayload({
+        taskOrder: ['coord-1'],
+        tasks: {
+          'coord-1': {
+            ...persistedTask(agentDef()),
+            id: 'coord-1',
+            coordinatorMode: true,
+          },
+        },
+        activeTaskId: 'coord-1',
+      }),
+    );
+
+    await loadState();
+
+    expect(store.tasks['coord-1']?.controlledBy).toBe('human');
+  });
+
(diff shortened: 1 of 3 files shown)


brooksc:
@FourWindff I agree with the concern here. I’m working on a broader follow-up that should make this PR unnecessary rather than layering more behavior onto the take/release control model.

The updated approach removes the need for users to manually take/release control in normal coordinator workflows. Prompts will be delivered automatically when the session is safe to receive them, and queued when there is real user activity: prompt drafts, pending terminal input, recent typing, or active questions/dialogs. Coordinated sub-task initial assignment delivery is also moving backend-side so background/unselected tasks can start reliably.

Once that lands, coordinator/user control should be much less visible and less error-prone.

brooksc:
@FourWindff I really appreciate you submitting this PR, I hope I didn't come across as discouraging you.  

I just submitted https://github.com/johannesjo/parallel-code/pull/149 which eliminates altogether the whole take/release control construct.  I was instead able to handle this smartly, detecting when the user is interacting and holding off on sending prompts - and detecting when it's idle.  This should make for a better UX without people having to learn this control feature.

I'd love your feedback on the coordinator as this is a new experimental feature.  Feel free to comment here or I started a discussion for coordinator questions/feedback/etc.  https://github.com/johannesjo/parallel-code/discussions/148

Thanks for trying this out!  And Kudos to @cledoux95 who started this feature!

FourWindff:
@brooksc Sorry for the late reply! I just tried out your latest PR and it looks great — much cleaner without the manual take/release control. The automatic detection of user activity is a much better UX.

  Let's continue any remaining discussion in the discussion thread so we can keep everything organized. Looking forward to more coordinator feedback once I've had more time with it!

PR #143, opened 2026-05-25
fix(worktree): prevent branch prefix conflicts

FourWindff:
## Summary

Prevents creating worktree branches whose prefix path conflicts with an existing local branch. For example, if branch `feature` already exists, the user can no longer create `feature/login/task-abc123`.

## Changes

- **Backend** (`electron/ipc/git.ts`): Add `findLocalBranchPrefixConflict` which checks all prefix segments of the target branch name against existing local branches via `git rev-parse` before running `git worktree add -b`.
- **Frontend** (`src/lib/branch-name.ts`): Add `findBranchPrefixConflict` to detect conflicts using the in-memory branch list, and `branchPrefixConflictError` for shared error message formatting.
- **UI** (`src/components/NewTaskDialog.tsx`): Wrap conflict detection in `createMemo` for efficiency; disable submit button and show inline error when a conflict exists.
- **UI** (`src/components/BranchPrefixField.tsx`): Accept `error` prop, render red border and error text.

## How it works

Git stores branches as ref files under `.git/refs/heads/`. A branch named `feature` creates a file at that path, which blocks any branch whose name starts with `feature/` (e.g. `feature/login`). This PR detects that condition both in the UI (early, with memoization) and at the backend (as a safety boundary).

## Test plan

- [x] Open New Task dialog for a git project
- [x] Set branch prefix to an existing branch name (e.g. `main`) → submit disabled, error shown
- [x] Set branch prefix to a path starting with existing branch (e.g. `main/task`) → submit disabled, error shown
- [x] Set a non-conflicting prefix → submit enabled
- [x] Backend returns clear error message if frontend check is bypassed

### DIFF
--- a/electron/ipc/git.ts
+++ b/electron/ipc/git.ts
@@ -218,6 +218,18 @@ async function localBranchExists(repoRoot: string, branch: string): Promise<bool
   }
 }
 
+async function findLocalBranchPrefixConflict(
+  repoRoot: string,
+  branchName: string,
+): Promise<string | null> {
+  const parts = branchName.split('/');
+  for (let i = 1; i < parts.length; i++) {
+    const prefix = parts.slice(0, i).join('/');
+    if (await localBranchExists(repoRoot, prefix)) return prefix;
+  }
+  return null;
+}
+
 async function detectMainBranchUncached(repoRoot: string): Promise<string> {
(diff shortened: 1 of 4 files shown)


johannesjo:
Thank you very much! <3

PR #155, opened 2026-05-29
feat(agents): add Antigravity CLI (agy) as built-in agent

### LINKED ISSUE #152: Antigravity CLI support
### :speaking_head: Foreword

Thank you for this amazing piece of software :) Would it be possible to support Antigravity CLI in the near future, since Gemini CLI will migrate to Antigravity CLI. 

## :person_frowning: Problem Statement

Antigravity CLI is not yet supported, and cannot be selected as a coding agent. 

## :grey_question: Possible Solution

Add support for Antigravity CLI

ASRagab:
## Add Antigravity CLI (`agy`) support

Fixes https://github.com/johannesjo/parallel-code/issues/152

Google is migrating Gemini CLI to the new **Antigravity CLI** (`agy`). On **June 18, 2026** Gemini CLI stops serving Google AI Pro/Ultra and free Code Assist users; it remains only on paid/enterprise API keys. This adds Antigravity as a first-class agent so users have a working migration path — **without removing Gemini CLI**, which paid/enterprise users still rely on. Fully additive, no breaking change.

### What changed
- **Agent registry** — new `antigravity` entry in `DEFAULT_AGENTS` (`electron/ipc/agents.ts`): `command: agy`, resume `-c`, skip-permissions `--dangerously-skip-permissions`, `prompt_ready_delay_ms: 1000`. Launched interactively like every other agent (prompt typed into the TUI, not passed as `-p`).
- **Docker (forward-compatible scaffolding only)** — `~/.gemini/antigravity-cli` added to the shared-auth mount map (`AGENT_CONFIG_DIRS`) for settings/plugins; `agy` installed in the bundled image via the official installer (Go binary, not npm). See the auth note below — **Antigravity is native-only**.
- **MCP guard** — `agy` excluded from both `--mcp-config` emission paths (`legacyMcpConfigArgs` in `src/lib/agent-args.ts` and `buildMcpLaunchArgs` in `electron/mcp/agent-args.ts`), the same way Codex is excluded. `agy` has no `--mcp-config` flag; passing it would break launch. **Scope consequence (intentional):** because `buildMcpLaunchArgs` returns nothing for `agy`, Antigravity tasks get **no** parallel-code MCP coordinator wiring — i.e. no spawning of parallel-code subagents from inside an `agy` session. Safe default for this change; wiring agy into the coordinator (it configures MCP via plugins/config, not a CLI flag) is a follow-up if wanted.
- **Gemini CLI untouched.**

### ⚠️ Antigravity is native-only (no Docker isolation)
- **Native (non-Docker) tasks work fully** via the normal interactive sign-in — credentials cached in the host OS keyring (Keychain/libsecret).
- **Docker-isolated Antigravity cannot authenticate.** Login is keyring-only OAuth; a barebones agent container has no secret-service daemon to reach the keyring, and `agy` has **no API-key environment fallback** (an earlier `ANTIGRAVITY_API_KEY` assumption was unverified and is false — `agy` still enters OAuth with it set). **Run Antigravity as a native task.**
- The Docker bits are kept as additive scaffolding: the bundled image ships the `agy` binary and `~/.gemini/antigravity-cli` (settings/plugins) is mounted when "Share agent auth" is enabled, so a future file-based / API-key auth path drops in cleanly.

### Corrections from Codex review
- Mount path fixed: `agy` writes app data to `~/.gemini/antigravity-cli`, **not** `~/.config/antigravity` (verified against the installed binary).
- Removed the unsupported `ANTIGRAVITY_API_KEY` Docker auth claim (binary does not consume it).
- `data/` runtime state (memory/sidecar `state_store.db` / `stream_store`) gitignored — it is generated local state, not source.

### Verification
- `tsc --noEmit` clean; full test suite green (111 tests across persistence, agents, both agent-args, pty). `pty.test.ts` config-dir-mount table extended with the corrected `agy` path.
- New unit tests assert no `--mcp-config` is emitted for `agy` on both arg paths.
- Manual native smoke test (selector availability, prompt delivery, `-c` resume, skip-permissions) and a full `docker build` are pending hardware/Docker.

CleanShot 2026-05-29 at 09 29 12

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ docs/
 /.ripgreprc
 /.zprofile
 /.zshrc
+
+# Local runtime state from memory/sidecar processes (not source)
+/data/
--- a/README.md
+++ b/README.md
@@ -115,9 +115,18 @@ When you're happy with the result, merge the branch back to main from the sideba
    - **macOS** — `.dmg` (universal)
    - **Linux** — `.AppImage` or `.deb`
 
-2. **Install at least one AI coding CLI:** [Claude Code](https://docs.anthropic.com/en/docs/claude-code), [Codex CLI](https://github.com/openai/codex), [Gemini CLI](https://github.com/google-gemini/gemini-cli), or [Copilot CLI](https://docs.github.com/en/copilot/concepts/agents/about-copilot-cli)
+2. **Install at least one AI coding CLI:** [Claude Code](https://docs.anthropic.com/en/docs/claude-code), [Codex CLI](https://github.com/openai/codex), [Gemini CLI](https://github.com/google-gemini/gemini-cli), [Antigravity CLI](https://antigravity.google/), or [Copilot CLI](https://docs.github.com/en/copilot/concepts/agents/about-copilot-cli)
 
 3. **Open Parallel Code**, point it at a git repo, and start dispatching tasks.
+
+<details>
+<summary><strong>Antigravity CLI: run natively, not in Docker isolation</strong></summary>
+
+Antigravity (`agy`) signs in interactively and caches credentials in your OS keyring (Keychain/libsecret). The keyring cannot be reached from inside a Linux container — it needs a secret-service daemon the agent container doesn't run — and `agy` has no API-key fallback, so **Docker-isolated Antigravity tasks cannot authenticate. Run Antigravity as a native (non-Docker) task.**
+
+The bundled image still ships the `agy` binary, and `~/.gemini/antigravity-cli` (settings/plugins) is shared when "Share agent auth" is enabled, so Docker support is forward-compatible if a file-based or API-key auth path appears later.
+
+</details>
 
 <details>
 <summary><strong>Build from source</strong></summary>
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -52,6 +52,14 @@ RUN ln -sf "$(command -v fdfind)" /usr/local/bin/fd 2>/dev/null || true
 # AI agent CLIs — must be present so Docker-mode tasks can execute them
 RUN npm install -g @anthropic-ai/claude-code @openai/codex @google/gemini-cli opencode-ai
 
+# Antigravity CLI (agy) — distributed as a Go binary via the official installer
+# (not on npm). The installer's `--dir` flag drops the binary straight into a
+# system path on PATH, so the non-root `agent` user that runs containers resolves
+# it. A failed fetch makes the build step fail, surfacing the break at image build
+# time rather than at task launch.
+RUN curl -fsSL https://antigravity.google/cli/install.sh | bash -s -- --dir /usr/local/bin \
+    && agy --version
+
 # Default git config so commits work out of the box inside containers
 RUN git config --system init.defaultBranch main \
     && git config --system advice.detachedHead false
--- a/electron/ipc/agents.ts
+++ b/electron/ipc/agents.ts
@@ -66,6 +66,17 @@ const DEFAULT_AGENTS: AgentDef[] = [
     // settle before sending, without being so long that the user notices the wait.
     prompt_ready_delay_ms: 1_000,
   },
+  {
+    id: 'antigravity',
+    name: 'Antigravity CLI',
+    command: 'agy',
+    args: [],
+    resume_args: ['-c'],
+    skip_permissions_args: ['--dangerously-skip-permissions'],
+    description: "Google's Antigravity CLI agent (successor to Gemini CLI)",
+    // Antigravity paints a TUI that needs a beat to settle before auto-send.
+    prompt_ready_delay_ms: 1_000,
+  },
 ];
 
 async function isCommandAvailable(command: string): Promise<boolean> {
--- a/electron/ipc/pty.test.ts
+++ b/electron/ipc/pty.test.ts
@@ -352,6 +352,7 @@ describe('spawnAgent docker mode', () => {
       ['gemini', '.gemini'],
       ['opencode', '.config/opencode'],
       ['copilot', '.config/github-copilot'],
+      ['agy', '.gemini/antigravity-cli'],
     ])(
       '%s bind-mounts a user-owned host directory when shareDockerAgentAuth is enabled',
       (command, relDir) => {
--- a/electron/ipc/pty.ts
+++ b/electron/ipc/pty.ts
@@ -710,6 +710,7 @@ const AGENT_CONFIG_DIRS: Record<string, string[]> = {
   gemini: ['.gemini'],
   opencode: ['.config/opencode'],
   copilot: ['.config/github-copilot'],
+  agy: ['.gemini/antigravity-cli'],
 };
 
 // Config files (not directories) each agent CLI uses for auth, relative to HOME.
--- a/electron/mcp/agent-args.test.ts
+++ b/electron/mcp/agent-args.test.ts
@@ -1,6 +1,11 @@
 import { describe, expect, it } from 'vitest';
 
-import { buildCodexMcpConfigOverride, buildMcpLaunchArgs, isCodexCommand } from './agent-args.js';
+import {
+  buildCodexMcpConfigOverride,
+  buildMcpLaunchArgs,
+  isAntigravityCommand,
+  isCodexCommand,
+} from './agent-args.js';
 
 const config = {
   mcpServers: {
@@ -49,5 +54,15 @@ describe('MCP agent launch args', () => {
       '--mcp-config',
       '/tmp/config.json',
     ]);
+  });
+
+  it('detects antigravity commands by executable name', () => {
+    expect(isAntigravityCommand('agy')).toBe(true);
+    expect(isAntigravityCommand('/home/agent/.local/bin/agy')).toBe(true);
+    expect(isAntigravityCommand('claude')).toBe(false);
+  });
+
+  it('emits no --mcp-config for Antigravity', () => {
+    expect(buildMcpLaunchArgs('agy', '/tmp/config.json', config)).toEqual([]);
   });
 });
--- a/electron/mcp/agent-args.ts
+++ b/electron/mcp/agent-args.ts
@@ -14,6 +14,10 @@ export function isCodexCommand(command: string): boolean {
   return command.split('/').pop()?.includes('codex') === true;
 }
 
+export function isAntigravityCommand(command: string): boolean {
+  return command.split('/').pop() === 'agy';
+}
+
 function tomlString(value: string): string {
   return JSON.stringify(value);
 }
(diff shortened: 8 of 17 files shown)


johannesjo:
Thank you ver ymuch for this @ASRagab <3!


## Multi-agent review

Reviewed by parallel Claude reviewers across correctness, security, architecture, alternatives, performance, and simplicity. The headline finding below was traced end to end through the PR-head code (including the backend `StartMCPServer` handler). Net: a clean, well-scoped, well-documented additive change with one issue worth addressing or consciously accepting before merge.

**Verdict: NEEDS CHANGES (minor).** Strong PR overall: faithful data-driven pattern, honest spec/docs (pending manual smoke tests correctly left unchecked), good coverage on both MCP-arg paths, accurate `agy` ground-truthing.

### Address or accept: `agy` coordinator/coordinated tasks flip to a false "MCP error" on restart

`src/store/tasks.ts` (`taskRequiresMcpLaunchArgs` / `applyTaskMcpLaunchResult`), reached from `App.tsx` rehydration and the retry button in `TerminalView.tsx`.

`agy` is the only agent for which `buildMcpLaunchArgs` returns `[]` while a `configPath` is still written. In `electron/ipc/register.ts` the coordinator `configPath` is written for any non-Docker coordinator regardless of agent, then `buildMcpLaunchArgs('agy', ...)` returns `[]` (Codex returns a non-empty `['--config', ...]`; all other agents return `['--mcp-config', ...]`).

Consequence:

- On initial create, a coordinator task is set `mcpStartupStatus: 'ready'` unconditionally, so an `agy` coordinator looks fine.
- On app restart (and on manual retry), `applyTaskMcpLaunchResult` receives `mcpLaunchArgs: []`. Because `Boolean(task.mcpConfigPath)` is true (persisted from create), `taskRequiresMcpLaunchArgs` returns true and the task is marked `markTaskMcpError('MCP startup returned no launch args')`.

So the same task reads `ready` on creation and `error` after every restart. It does not crash or lose data, and it only bites if `agy` is used in coordinator/coordinated mode (documented as having no MCP wiring), but nothing in the UI prevents selecting `agy` as a coordinator, so the path is reachable.

Note this is also a third copy of command detection: `src/store/tasks.ts` has its own `isCodexCommand` that was never updated for `agy`, which is why the site was missed.

Clean fix (small, in spirit): make `taskRequiresMcpLaunchArgs` treat `agy` as not requiring launch args, e.g. `isCodexCommand(cmd) || (Boolean(task.mcpConfigPath) && !isAntigravityCommand(cmd))`. Alternatives: gate coordinator mode off for `agy`, or do not persist `mcpConfigPath` for it. This lives in coordinator infrastructure this PR did not author, so scoping it as a follow-up is reasonable, but the `ready`-then-`error` inconsistency is real today.

### Other warnings

- **Dockerfile `curl | bash` installer is unpinned (`docker/Dockerfile`).** `curl -fsSL https://antigravity.google/cli/install.sh | bash -s -- --dir /usr/local/bin && agy --version` pipes a remote, unversioned script into root at build time with no checksum or signature; `&& agy --version` proves a binary installed, not which one. The `npm install -g` line above is also unpinned, but the repo's other network fetches (NodeSource, GitHub CLI) use GPG-verified apt keyrings, so this breaks that pattern. Blast radius is bounded (Docker only, and `agy` is documented native-only so the bundled binary is currently unused). Suggest pinning a versioned release plus a published SHA256 if upstream supports it, or at minimum documenting the supply-chain risk in design.md (it currently flags only reproducibility).
- **`openspec/changes/add-antigravity-agent/.openspec.yaml` is inconsistent with the repo.** None of the existing change folders carry a `.openspec.yaml`. Harmless (validation passes either way), but suggest dropping it unless adopting repo-wide.

### Suggestions (optional)

- Consolidate the duplicated `isCodexCommand` / `isAntigravityCommand` detection (now in 3-4 places) and use the currently-dead `AgentDef.mcp_config_flag` field, or one shared module, as the source of truth. This is what would have caught the issue above. Pre-existing pattern and larger than this PR, so a reasonable follow-up.
- The `agy` coordinator preamble routes into the `else` (Claude) branch in `coordinator.ts` and writes to `.claude/settings.local.json`, which `agy` does not read (its config is `~/.gemini/antigravity-cli`). Silent no-op in the unsupported coordinator path. Same root cause as above.
- Consider a one-line "why" comment on `isAntigravityCommand`'s exact match (vs Codex's substring) and on the `agy: ['.gemini/antigravity-cli']` nesting under `gemini: ['.gemini']` in `AGENT_CONFIG_DIRS`, so neither gets "fixed" later.

### Checked and cleared (not issues)

- **No credential leak via the `agy` mount.** The mount source is an isolated, fresh `0o700` per-agent dir (`~/.parallel-code/agent-auth/agy/...`), not the host's real `~/.gemini`, and it is gated behind the share-auth opt-in. Verified against `pty.test.ts`.
- **`isAntigravityCommand` exact match vs `isCodexCommand` substring is fine.** Exact match is the safer choice for a short token like `agy`; not a defect.
- **No meaningful performance impact.** One extra TTL-cached `which` probe, one keyed selector row, no renderer-bundle delta, build-time-only Docker cost.
- **`AgentDef` duplication** (`src/ipc/types.ts` vs `electron/ipc/agents.ts`) is pre-existing and correctly left untouched (no new field added).

Note: a seventh reviewer (Codex CLI) could not run in this environment (read-only filesystem blocked its app-server init), so this summary reflects the six Claude reviewers plus manual verification.

ASRagab:
> Dockerfile curl | bash installer is unpinned (docker/Dockerfile). curl -fsSL https://antigravity.google/cli/install.sh | bash -s -- --dir /usr/local/bin && agy --version pipes a remote, unversioned script into root at build time with no checksum or signature; && agy --version proves a binary installed, not which one.

This is overly precious I think, the other clis are also not pinned.

johannesjo:
## Multi-agent review — PR #155 (add Antigravity CLI `agy`)

Reviewed by 7 independent agents (Claude × 6: correctness, security, architecture, alternatives, performance, simplicity; + Codex CLI), with each finding verified against the actual code. The diff reviewed is the precise 16-file PR scope (459/-4), isolated from the stacked #149 base.

**Verdict: Approve — no blocking issues.** The change is correct, well-scoped, well-tested, and unusually well-documented. Findings below are all non-blocking quality/maintainability notes. Two reviewer claims were checked and dismissed (see end).

### ✅ What holds up (verified)
- **`taskRequiresMcpLaunchArgs` change is a genuine, necessary fix**, not a workaround. Traced the truth table: it returns `true` for Codex always, `false` for `agy` (which legitimately emits `[]`), and is unchanged for every other agent. Confirmed via spawn-gating (`TerminalView.tsx`: agent spawns only when `mcpStartupStatus === 'ready'`) that **without** this change an `agy` coordinator task would be marked `error` and never launch — i.e. agy coordinators would be unusable. The missing MCP coordinator wiring is the documented, accepted trade-off.
- **Both `--mcp-config` emission paths exclude `agy`** (`legacyMcpConfigArgs`, `buildMcpLaunchArgs`) — these are the only two production sites that emit the flag, so no path can hand `agy` an unsupported flag.
- **Docker auth mount is safe**: the bind source is the sandboxed per-agent dir `~/.parallel-code/agent-auth/agy/.gemini/antigravity-cli`, *not* the host's real `~/.gemini` — so Gemini credentials are not exposed and there's no cross-agent bleed.
- **Performance-neutral**: one extra parallel `which agy` probe (TTL-cached), one extra keyed `<For>` button. No hot-path cost.

### 🟡 Worth considering (non-blocking)

**1. Command-sniffing duplication + dead `mcp_config_flag` field** *(flagged by architecture, alternatives, correctness, simplicity)*
`isAntigravityCommand` and `isCodexCommand` are each now hand-copied into **3 files** (`src/lib/agent-args.ts`, `electron/mcp/agent-args.ts`, `src/store/tasks.ts`) — 6 near-identical predicates that must stay in lockstep, with subtly different signatures (`string` vs `string | undefined`) and matching strategies (`=== 'agy'` vs `.includes('codex')`). Meanwhile `AgentDef.mcp_config_flag` (`src/ipc/types.ts:21`) is **declared but read by zero call sites** (pre-existing in `main`, not introduced here).

This mirrors the existing Codex pattern, so shipping as-is is defensible. But the threshold is now clearly crossed. Two options for a follow-up:
- **Minimal**: hoist the predicates to one shared module (`src/` already imports from `electron/` elsewhere, so the IPC boundary isn't a blocker).
- **Better**: make MCP-arg behavior data-driven via a per-agent capability (e.g. `mcp_config_mode?: 'flag' | 'codex' | 'none'`) resolved through an `agents.ts` helper mirroring the existing `getSkipPermissionsArgs`, then delete all 6 predicates. This is what `design.md` *implies* already exists.

Either way: **`design.md` Decision #4 should be corrected** — it leans on `mcp_config_flag` as if it were a live declarative opt-out, but exclusion is actually imperative command-name matching. (Smaller scope than a refactor; worth fixing the doc regardless.)

**2. `curl … install.sh | bash` is the least-protected fetch in the Dockerfile** *(security, alternatives, simplicity)*
```dockerfile
RUN curl -fsSL https://antigravity.google/cli/install.sh | bash -s -- --dir /usr/local/bin && agy --version
```
This runs arbitrary remote code as root at build time with no pin or signature, while the *other* external fetches in the same file (NodeSource, gh) install via signed GPG apt repos. It's consistent with the already-unpinned `npm install -g …` line and acceptable for a latest-tracking bundled image, so this is a conscious-decision flag, not a defect. Cheap hardening if you want it: **pin the known-good `agy 1.0.3`** (a verified-good version already exists per design.md). Also: design.md's "the upstream installer verifies the release payload checksum" overstates the protection — that checksum lives *inside* the unpinned script, so it's not an independent guarantee.

**3. `taskRequiresMcpLaunchArgs` readability** *(architecture, alternatives)* — the `A || (B && !C)` form is correct but makes the agy carve-out hard to read. An early return matches the function's own existing guard style and is behavior-identical:
```ts
const cmd = agentDef?.command;
if (isAntigravityCommand(cmd)) return false; // agy legitimately emits no MCP args
return isCodexCommand(cmd) || Boolean(task.mcpConfigPath);
```

### 🔵 Minor / optional
- Switch only the **electron** copy of the basename parse to `path.basename(...)` (the repo's idiom there: `pty.ts`, `agents.ts`); leave the renderer copies on `.split('/').pop()` (the renderer's idiom). *(alternatives)*
- Add a one-line comment explaining why `agy` uses exact-match (`=== 'agy'`) vs Codex's substring match — saves a future reader the "why the asymmetry?" detour. *(correctness, simplicity)*

### ⚠️ Reviewer claims checked and dismissed
- **"Remove `&& agy --version` — `curl -f` already fails the build"** — **incorrect, keep it.** The Dockerfile sets no `pipefail` (the `ENV SHELL=/bin/bash` is a runtime var, not the build `RUN` shell, which is `/bin/sh -c`). So in `curl -fsSL … | bash`, a curl failure is masked and the pipeline exits 0. The trailing `agy --version` is the **only** thing that deterministically fails the build on a bad fetch — it's load-bearing.
- **"The `~/.gemini` mount could expose Gemini credentials"** — **not realized.** The mount uses an isolated per-agent sandbox dir (see verified section above).

---
*Agreement: 0 of 7 reviewers found a CRITICAL or blocking issue; Codex (constrained to the diff) reported no findings. The DRY/`mcp_config_flag` theme was independently raised by 4 of the 6 Claude reviewers.*

🤖 Generated with multi-agent review (Claude Code)

johannesjo:
Thank you very much for this <3

PR #164, opened 2026-06-01
feat(ask-code): upgrade MiniMax model to M3

octo-patch:
## Summary

Bumps the bundled MiniMax model id from `MiniMax-M2.7` to `MiniMax-M3` in the inline code Q&A backend (`electron/ipc/ask-code-minimax.ts`). M3 is the current Minimax model and is a drop-in replacement on the OpenAI-compatible endpoint at `api.minimax.io` — same 204K context window, same request schema, same temperature constraints, same `MINIMAX_API_KEY` env var, so the rest of the provider code is unchanged.

## Changes

- `electron/ipc/ask-code-minimax.ts` — `MINIMAX_MODEL` constant flipped to `MiniMax-M3`.
- `electron/ipc/ask-code-minimax.test.ts` — model-assertion test renamed to match (still asserts against the exported `MINIMAX_MODEL`, so the test stays semantically equivalent).
- `src/components/SettingsDialog.tsx` — provider dropdown option and the explainer copy below it now read "M3" instead of "M2.7".
- `README.md` — feature blurb in "Why Parallel Code?" mentions M3 instead of M2.7.

No changes to the API URL, temperature range, streaming flow, API key handling, or any non-minimax provider.

## Test plan

- [x] `npm run typecheck` — clean.
- [x] `npm run lint` — clean.
- [x] `npm run format:check` — clean.
- [x] `npx vitest run electron/ipc/ask-code-minimax.test.ts` — 12/12 pass (including the renamed model-id assertion and the temperature-range assertion).

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@
 - **See every session in one place** — switch context without losing momentum.
 - **Control everything keyboard-first** — every action has a shortcut, mouse optional.
 - **Monitor progress from your phone** — scan a QR code, watch agents work over Wi-Fi or Tailscale.
-- **Ask about code with any LLM** — the inline code Q&A feature supports [Claude Code](https://docs.anthropic.com/en/docs/claude-code) (default) or [MiniMax](https://www.minimax.io/) M2.7 (204K context) — configurable in Settings.
+- **Ask about code with any LLM** — the inline code Q&A feature supports [Claude Code](https://docs.anthropic.com/en/docs/claude-code) (default) or [MiniMax](https://www.minimax.io/) M3 (204K context) — configurable in Settings.
 
 <details>
 <summary><strong>How does it compare?</strong></summary>
--- a/electron/ipc/ask-code-minimax.test.ts
+++ b/electron/ipc/ask-code-minimax.test.ts
@@ -164,7 +164,7 @@ describe('askAboutCodeMinimax', () => {
     );
   });
 
-  it('uses MiniMax-M2.7 model', async () => {
+  it('uses MiniMax-M3 model', async () => {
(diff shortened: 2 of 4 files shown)


johannesjo:
**Reviewed — the M3 bump is correct and safe to ship, with one copy fix I'd make first.**

✅ **The core change holds up.** I verified `MiniMax-M3` against MiniMax's published docs: it's the current model (released Jun 1, 2026), served on the same `api.minimax.io/v1/chat/completions` endpoint with the same bearer-auth and OpenAI-compatible schema this code already uses. The "drop-in replacement" claim is accurate for the request flow.

✅ **Complete.** Grepping the tree for `M2.7` turns up only the four files this PR touches; `PRIVACY.md` references MiniMax but carries no version string, so nothing was missed.

⚠️ **The "204K context" copy is now inaccurate.** The PR description says M3 has the "same 204K context window," but M3's published spec is **up to 1M tokens (min 512K guaranteed)** — a headline M3 feature, not an M2.7 carryover. As written, `README.md:55` and `SettingsDialog.tsx:519,554` would advertise "M3 (204K context)" and *undersell* the model you're upgrading to. Suggest updating the parenthetical (e.g. "1M context") in this same PR.

ℹ️ **Minor, pre-existing:** the renamed `uses MiniMax-M3 model` test asserts `body.model === MINIMAX_MODEL` — i.e. it compares the payload against the same constant used to build it. It guards that the constant flows through, but can't catch a wrong/typo'd model id, so the rename is cosmetic. Not introduced here; just noting the test name promises more than it checks.

**Verdict:** Approve once the context-window copy is updated; functionally good as-is.

PR #165, opened 2026-06-02
feat(settings): add new task defaults for steps, skip-permissions, and propagate

brooksc:
## Problem

Every time a user creates a new task, three commonly-toggled options reset to
their hardcoded defaults:

- **Steps tracking** always starts unchecked
- **Dangerously skip all confirms** always starts unchecked
- **Propagate skip-permissions to sub-tasks** (coordinator mode) always starts unchecked

Users who habitually enable one or more of these options have to manually
re-check them for every single task, which is repetitive friction.

## Solution

Add three persistent settings under a new **"New Task Defaults"** section in
**Settings → General**:

| Setting | What it does |
|---|---|
| **Steps tracking** | Pre-checks "Steps tracking" in the New Task Dialog |
| **Dangerously skip all confirms** | Pre-checks "Dangerously skip all confirms" (when the selected agent supports it) |
| **Propagate skip-permissions to sub-tasks** | Pre-checks propagate in coordinator mode (only shown when Coordinator mode is enabled in Settings) |

The propagate option is conditionally rendered only when `coordinatorModeEnabled`
is true — matching the same conditional in the New Task Dialog itself.

## Implementation

**Store layer** (`src/store/`)
- `types.ts` — added `defaultStepsEnabled`, `defaultSkipPermissions`,
  `defaultPropagateSkipPermissions` to both `AppStore` and `PersistedState`
- `core.ts` — default values (all `false`)
- `ui.ts` — three setter functions following the existing pattern
- `store.ts` — re-exports the new setters
- `persistence.ts` — save/load with the standard falsy-omit pattern
  (field absent from JSON when `false`, explicit `true` when set); also
  added to `LegacyPersistedState` for forward-compat parsing

**UI layer** (`src/components/`)
- `SettingsDialog.tsx` — new "New Task Defaults" section inserted in the
  General tab, before "Ask about Code"; three checkbox rows matching the
  existing label+description card style
- `NewTaskDialog.tsx` — `stepsEnabled`, `skipPermissions`, and
  `propagateSkipPermissions` signals now initialise from the store defaults
  instead of hardcoded `false`/`store.showSteps`

## Tests

44 new test cases added to `src/store/persistence.test.ts`, one describe
block per setting (`new task defaults persistence`). Each setting gets five
coverage points:

1. **Defaults to `false`** when the field is absent from saved state
2. **Restores `true`** from saved state
3. **Omits `false`** from the serialised JSON (falsy-omit pattern)
4. **Persists `true`** through `saveState`
5. **Rejects non-booleans** (string, number, null) and falls back to `false`

All 1113 tests pass on this branch (full suite: `npm test`).


johannesjo:
Thank you very much! <3

## Review summary

Nice clean store-layer plumbing and thorough save/load tests. Two blockers before merge, plus some nits.

---

### 🛑 Blocker 1 — Autosave never fires for these toggles

The new setters in `src/store/ui.ts:215-225` only call `setStore(...)` and rely on the autosave effect. But `persistedSnapshot()` in `src/store/autosave.ts:8-87` does **not** include the three new fields. The effect compares stringified snapshots, so toggling a default in Settings produces no diff, never schedules `saveState()`, and the change is lost if no other tracked field mutates before quit.

The `persistence.test.ts` tests pass because they call `saveState()` / `loadState()` directly and bypass the autosave path entirely.

**Fix:** add three lines to `persistedSnapshot()` (alongside `coordinatorModeEnabled` at line 38):

```ts
defaultStepsEnabled: store.defaultStepsEnabled,
defaultSkipPermissions: store.defaultSkipPermissions,
defaultPropagateSkipPermissions: store.defaultPropagateSkipPermissions,
```

Worth adding an autosave-level regression test too — this same footgun would catch any future persisted-field PR.

---

### 🛑 Blocker 2 — Breaks documented sticky-default behavior for existing users

`store.showSteps` is misnamed: nothing reads it for rendering. It's purely the sticky last-used default for the new-task dialog, codified in `openspec/specs/steps-tracking/spec.md` (lines ~11-26: dialog pre-checked iff `showSteps` true; toggling updates `showSteps`). The new dialog default at `src/components/NewTaskDialog.tsx:69` switches from `store.showSteps` → `store.defaultStepsEnabled`, breaking both scenarios:

1. **Existing users with `showSteps: true` lose their preference** on first launch post-PR — `loadState` (`src/store/persistence.ts:608-610`) reads only `raw.defaultStepsEnabled === true`.
2. **`tasks.ts:247-249` writeback is now orphaned** — the dialog always passes `opts.stepsEnabled` explicitly, so `opts.stepsEnabled ?? store.showSteps` at line 247 is unreachable from the dialog and the write-back at line 249 mutates a value nothing else reads.
3. **No `openspec/changes/<name>/` delta** — per `CLAUDE.md`, new/changed behavior needs a proposal; this PR also doesn't update `specs/steps-tracking/`.

Pick one:
- **(a)** Migrate: in `loadState`, `s.defaultStepsEnabled = raw.defaultStepsEnabled === true || raw.showSteps === true;` then remove `showSteps` from store + persistence + `tasks.ts:247-249`, and update the spec.
- **(b)** Drop `defaultStepsEnabled`, keep `showSteps` as source of truth (only add the two new fields).

Either way, add the OpenSpec change.

---

### Nits (non-blocking)

- **Propagate gating mismatch.** Settings hides the row behind `store.coordinatorModeEnabled` only (`SettingsDialog.tsx:540`), but the dialog's checkbox is gated by `coordinatorModeEnabled` → `coordinatorMode()` → `agentSupportsSkipPermissions() && skipPermissions()` (`NewTaskDialog.tsx:1167/1202/1247`). The Settings help text "Default propagate-skip-permissions on when coordinator mode is enabled" is misleading — the default only takes effect when the user also enables coord mode *and* skip-perms on a given task.
- **Description copy doesn't match the existing voice** in this file (compare to "Show native notifications when tasks finish or need attention", "Display Claude Code plan files in a tab next to Notes"). The new ones restate the title. Suggest:
  - Steps: "Pre-tick Steps tracking in the New Task dialog."
  - Skip-permissions: "Pre-tick skip-permissions in the New Task dialog (only honored when the selected agent supports it)."
  - Propagate: "Pre-tick Propagate to sub-tasks when both coordinator mode and skip-permissions are enabled for a task."
- **Title voice mismatch.** "Dangerously skip all confirms" mixes voice with "Steps tracking" and "Propagate skip-permissions to sub-tasks". Consider "Skip permissions" (matches the store key and `NewTaskDialog` terminology).
- **Placement.** Consider folding into / adjacent to the existing `Behavior` group (`SettingsDialog.tsx:225`) rather than between `Editor` and `Ask about Code` — these are behavior defaults.
- **Test count.** PR description says 44 tests; diff shows 21 (7 per setting). Cosmetic.

Inline-style duplication and missing `aria-*`/`for` follow the existing convention in this file — not regressions, separate cleanup if anything.

brooksc:
Thanks for the thorough review — all issues addressed in the latest push. Here's what was fixed, in the same order as your comments:

---

### 🛑 Blocker 1 — Autosave never fires

**Root cause confirmed:** `persistedSnapshot()` in `autosave.ts` did not include the three new fields, so toggling them in Settings produced no snapshot diff and `saveState()` was never scheduled.

**Fix:** Added all three fields to `persistedSnapshot()`:
```ts
defaultStepsEnabled: store.defaultStepsEnabled,
defaultSkipPermissions: store.defaultSkipPermissions,
defaultPropagateSkipPermissions: store.defaultPropagateSkipPermissions,
```

**Regression test:** Added `autosave.test.ts` with four cases — one per field (verifies toggling produces a snapshot diff) plus a guard that `showSteps` no longer exists on the store, so any future re-addition without updating `autosave.ts` fails loudly.

---

### 🛑 Blocker 2 — Breaks existing users' `showSteps` preference

**Fix chosen: option (a)** — migrate `showSteps` into `defaultStepsEnabled` and retire the old key.

- `loadState` now reads: `s.defaultStepsEnabled = raw.defaultStepsEnabled === true || raw.showSteps === true` — existing users with `showSteps: true` keep their preference transparently.
- Removed `showSteps` from `AppStore`, `PersistedState`, `core.ts`, and `saveState()` output. Kept it in `LegacyPersistedState` (typed `unknown`) for forward-compat parsing.
- Both write-backs in `tasks.ts` updated: the `createTask` fallback (`opts.stepsEnabled ?? store.showSteps`) and `setTaskStepsEnabled` both now use `defaultStepsEnabled`.
- **4 migration tests** added to `persistence.test.ts`: migrate `showSteps=true`, new field wins on conflict, `showSteps=false` stays false, `saveState` no longer writes `showSteps`.
- **OpenSpec change** added at `openspec/changes/new-task-defaults/change.md` documenting the `showSteps` rename and the two new fields.

---

### Nits

- **Autosave regression test** — covered above.
- **Propagate gating description** — updated to: *"Pre-tick Propagate to sub-tasks when both coordinator mode and skip-permissions are enabled for a task"* — accurately reflects that the default only applies when both conditions are met.
- **Description copy** — all three rewritten to match the existing voice:
  - Steps: *"Pre-tick Steps tracking in the New Task dialog"*
  - Skip: *"Pre-tick skip-permissions in the New Task dialog (only honoured when the selected agent supports it)"*
  - Propagate: *"Pre-tick Propagate to sub-tasks when both coordinator mode and skip-permissions are enabled for a task"*
- **Title voice** — "Dangerously skip all confirms" → **"Skip permissions"**
- **Placement** — "New Task Defaults" section moved to sit immediately after the "Behavior" group (before Editor), where it fits naturally as behavior defaults.
- **Test count** — cosmetic only, no change needed.

johannesjo:
## Review summary — needs changes 🔴

Thanks for this! The store/persistence/UI wiring is clean, consistent, and well-covered at the persistence layer. However, **the feature doesn't actually work for 2 of its 3 settings**, because the New Task dialog never consumes the values it persists. Details below, ordered by priority.

### 🔴 Critical — `Skip permissions` and `Propagate` defaults have no effect

`src/components/NewTaskDialog.tsx`

The PR points the signal initializers at the store:

```ts
const [skipPermissions, setSkipPermissions] = createSignal(store.defaultSkipPermissions);            // ~L69
const [propagateSkipPermissions, setPropagateSkipPermissions] =
  createSignal(store.defaultPropagateSkipPermissions);                                               // ~L81
```

…but the pre-existing "initialize on open" effect (unchanged by this PR) immediately forces both back to `false` every time the dialog opens:

```ts
createEffect(() => {
  if (!props.open) return;
  ...
  setSkipPermissions(false);            // ~L161 — overrides the new default
  setPropagateSkipPermissions(false);   // ~L162 — overrides the new default
```

So those two initializers are dead code and the checkboxes are always unchecked on open — contradicting this change's own `change.md` ("pre-checked iff `defaultSkipPermissions` is true").

There's also a related, **pre-existing** issue for steps (so not a regression, but it means the steps default is unreliable too): `NewTaskDialog` is rendered unconditionally, so `createSignal(store.defaultStepsEnabled)` runs during the initial render, **before** `loadState()` runs in `onMount` (`App.tsx:385`). The non-reactive initial value can capture `false` before the persisted value lands.

**Fix — set all three from the store inside the open effect** (which runs after `loadState`, so the store is populated). This resolves both problems at once:

```ts
setStepsEnabled(store.defaultStepsEnabled);
setSkipPermissions(store.defaultSkipPermissions);
setPropagateSkipPermissions(store.defaultPropagateSkipPermissions);
```

A component/e2e test asserting the checkbox `checked` state after load + open would have caught this and is worth adding.

### 🟡 `src/store/autosave.test.ts` is a tautology (false confidence)

It re-implements the snapshot inline (`captureSnapshot()`) because the real `persistedSnapshot()` isn't exported, then asserts `JSON.stringify({x:false}) !== JSON.stringify({x:true})` — true by construction. The comment claims "if autosave.ts diverges … tests catch it," but deleting a field from the real `persistedSnapshot()` (`autosave.ts:28`) leaves every test green. Only the `'showSteps' in store === false` case has value. **Recommend** exporting `persistedSnapshot` and testing it directly, or deleting the file (the `persistence.test.ts` round-trips already cover the real behavior).

### 🟡 OpenSpec change doesn't follow the repo workflow

`openspec/changes/new-task-defaults/` contains only `change.md`; every other change uses `proposal.md` + `tasks.md` + a `specs/<capability>/spec.md` delta. I ran the tooling: the folder is **invisible** to `openspec validate --all --strict` (no `proposal.md`), so it doesn't *fail* the gate — but it also means the `showSteps → defaultStepsEnabled` rename ships **no spec delta**, and `openspec/specs/steps-tracking/spec.md` still documents the old `showSteps` flag. Please add `proposal.md` / `tasks.md` and a `MODIFIED Requirements` delta for `steps-tracking` so the rename reconciles on archive.

### 🟢 Minor / nice-to-have

- **Dead type:** `PersistedState.showSteps` (`types.ts:246`) is now never read or written — the migration reads from `LegacyPersistedState`. Remove it, plus the stray blank line left at `persistence.ts:135`.
- **Security/UX labeling:** the Settings row says "Skip permissions"; the dialog says "**Dangerously** skip all confirms" with a warning. Since this makes confirmation-bypass sticky for every future task, consider mirroring the stronger wording in Settings. (The actual gating is sound — `agentSupportsSkipPermissions() && skipPermissions()` — and the danger banners stay reactive, so there's no vulnerability.)
- **DRY (optional):** the 3 new checkbox cards repeat a ~20-line block already duplicated 13× in `SettingsDialog.tsx`; a small local `SettingToggle` component would help. Fine to scope to the new rows or defer.
- **Tests (optional):** the three structurally-identical per-field persistence blocks could collapse into a `describe.each`; the non-boolean cases are near-tautologies given the `=== true` loader. Keep the migration block — that part earns its keep.

### Not issues (verified)
- Persistence falsy-omit round-trip, strict `=== true` validation, and the `showSteps` migration are correct and can't be tricked into enabling a dangerous default via malformed state.
- No performance concerns (no redundant autosave writes, O(1) persistence path).
- The store-layer repetition (one-line setters/re-exports) correctly matches the codebase convention — no abstraction needed there.

---
*Reviewed with a 7-agent parallel pass (correctness, security, architecture, alternatives, performance, simplicity, + Codex CLI); all headline findings verified against the code.*

brooksc:
## Review summary — addressed ✅

Thanks for the thorough second pass — all issues addressed in the latest push. Here's what was fixed, in the same order as your comments:

---

### 🔴 Critical — `Skip permissions` and `Propagate` defaults have no effect

**Root cause confirmed:** The pre-existing open effect (`NewTaskDialog.tsx:153`) reset `skipPermissions` and `propagateSkipPermissions` to `false` on every open, making the store-based signal initializers dead code. Additionally, `createSignal(store.defaultStepsEnabled)` captured the store value during the initial render, before `loadState()` ran in `onMount`, so the steps default was also unreliable.

**Fix:** All three signals are now set from the store *inside* the open effect (which runs after `loadState`), resolving both the override and the race condition at once:
```ts
setStepsEnabled(store.defaultStepsEnabled);
setSkipPermissions(store.defaultSkipPermissions);
setPropagateSkipPermissions(store.defaultPropagateSkipPermissions);
```
(`NewTaskDialog.tsx:161-164`)

---

### 🟡 `src/store/autosave.test.ts` is a tautology (false confidence)

**Root cause confirmed:** `persistedSnapshot()` was unexported, so the test re-implemented it inline — any drift in the real function left every test green.

**Fix:** Exported `persistedSnapshot` from `autosave.ts` and rewrote the test to import and call it directly. The three field-change tests now exercise the real snapshot function, so deleting a field from `autosave.ts` will break the corresponding test. The `showSteps not in store` guard is retained.

---

### 🟡 OpenSpec change doesn't follow the repo workflow

**Root cause confirmed:** The `openspec/changes/new-task-defaults/` folder contained only `change.md` with no `proposal.md`, `tasks.md`, or `specs/` delta, making it invisible to `openspec validate --all --strict` and leaving `specs/steps-tracking/spec.md` documenting the old `showSteps` flag.

**Fix:** Added `proposal.md`, `tasks.md`, and `specs/steps-tracking/spec.md` with `## MODIFIED Requirements` (renaming `showSteps` → `defaultStepsEnabled`) and `## ADDED Requirements` (the two new fields). `openspec validate --all --strict` now passes for this change. The canonical `openspec/specs/steps-tracking/spec.md` is also updated to replace `showSteps` with `defaultStepsEnabled` in both scenarios.

---

### 🟢 Minor / nice-to-have

- **Dead type** — `PersistedState.showSteps` removed from `types.ts`. Stray blank line at `persistence.ts:135` removed.
- **Security/UX labeling** — no change; leaving wording as-is for now. The setting controls a default, not a permanent bypass, and the danger banner in the dialog remains reactive and visible when the default fires. Happy to revisit if you feel strongly.
- **DRY (optional)** — deferred; the duplication predates this PR and a `SettingToggle` refactor is a separate cleanup.
- **Tests (optional)** — deferred; the `describe.each` consolidation is a cosmetic improvement best done in isolation.

johannesjo:
Thanks for the detailed PR and tests — the migration story and OpenSpec proposal are great. A few correctness issues to address before merging:

**High — reset effect tracks store defaults, can clobber in-flight edits** (`src/components/NewTaskDialog.tsx:164`)
The dialog's open-effect reads `store.defaultStepsEnabled`, `store.defaultSkipPermissions`, and `store.defaultPropagateSkipPermissions` while `props.open` is true. Solid registers those as reactive dependencies of the whole effect, so any later write to any of them — toggling the Settings row (both dialogs are rendered simultaneously in `App.tsx`), `loadState` re-running, autosave restoration — re-fires the entire "fresh dialog" block and wipes `prompt`, `name`, `selectedAgent`, drop URL, and the rest. Sample the defaults only on the `props.open` transition, e.g.

```tsx
createEffect(on(() => props.open, (open) => {
  if (!open) return;
  untrack(() => {
    setStepsEnabled(store.defaultStepsEnabled);
    setSkipPermissions(store.defaultSkipPermissions);
    setPropagateSkipPermissions(store.defaultPropagateSkipPermissions);
  });
  // … rest of reset
}));
```

**Medium — hidden `propagateSkipPermissions: true` leaks into task metadata** (`src/components/NewTaskDialog.tsx:632`)
Submission gates only on `coordinatorMode()`:
```ts
propagateSkipPermissions: coordinatorMode() ? propagateSkipPermissions() : undefined,
```
But the checkbox UI is gated on `agentSupportsSkipPermissions() && skipPermissions()` (line 1269). With `defaultPropagateSkipPermissions=true`, a coordinator task on an agent that doesn't support skip-permissions (or with skip-permissions unchecked) silently ships `propagateSkipPermissions: true` to the backend — the user never sees the checkbox. Gate submission the same way the UI gates visibility:
```ts
propagateSkipPermissions:
  coordinatorMode() && agentSupportsSkipPermissions() && skipPermissions()
    ? propagateSkipPermissions()
    : undefined,
```

**Low — legacy `showSteps` overrides explicit new key** (`src/store/persistence.ts:577`)
```ts
s.defaultStepsEnabled = raw.defaultStepsEnabled === true || raw.showSteps === true;
```
If a saved state has `defaultStepsEnabled: false` and `showSteps: true` (possible during partial migration or hand-edited state), the explicit `false` is overridden. Prefer presence-checking the new field:
```ts
s.defaultStepsEnabled =
  typeof raw.defaultStepsEnabled === 'boolean'
    ? raw.defaultStepsEnabled
    : raw.showSteps === true;
```
The existing "new field wins" test only covers the `true/false` direction — adding the `false/true` case would catch this.

**Low — Settings row should mirror the danger wording** (`src/components/SettingsDialog.tsx:466`)
The per-task checkbox is labelled **"Dangerously skip all confirms"**, but the persistent default reads neutrally as **"Skip permissions"**. The default is the higher-stakes setting (auto-enables for every future task), so it should carry the same warning framing — "Dangerously skip all confirms by default" or similar, with a description noting the blast radius.

**Also**
- `src/store/tasks.ts:1582` reintroduces `setTaskStepsEnabled` with zero callers (flagged by knip). It was deliberately removed in `33b2e4f` for being dead code — please drop it again.

brooksc:
All issues addressed in the latest push, in order:

---

**🔴 High — open effect re-fires on store mutations**

Confirmed: reading `store.defaultStepsEnabled` etc. directly inside `createEffect()` registered them as reactive dependencies, so any Settings toggle or autosave restore while the dialog was open re-fired the entire reset block.

Fix: wrapped in `createEffect(on(() => props.open, ..., { defer: true }))` and moved the three default reads inside `untrack()`, so they are one-shot samples taken only on the open transition — not ongoing reactive subscriptions.

---

**🟡 Medium — `propagateSkipPermissions` leaked into task metadata when hidden**

Confirmed: submission only gated on `coordinatorMode()` but the checkbox is only visible when `agentSupportsSkipPermissions() && skipPermissions()` are also true.

Fix: gate submission the same way the UI gates visibility:
```ts
propagateSkipPermissions:
  coordinatorMode() && agentSupportsSkipPermissions() && skipPermissions()
    ? propagateSkipPermissions()
    : undefined,
```

---

**🟡 Low — legacy `showSteps` overrode explicit `defaultStepsEnabled:false`**

Confirmed: `raw.defaultStepsEnabled === true || raw.showSteps === true` treated the legacy field as additive, not as a fallback.

Fix: presence-check the new field first, fall back to `showSteps` only when absent:
```ts
s.defaultStepsEnabled =
  typeof raw.defaultStepsEnabled === 'boolean'
    ? raw.defaultStepsEnabled
    : raw.showSteps === true;
```

Added the missing reverse test (`showSteps:true + defaultStepsEnabled:false → result is false`) to lock in both directions.

---

**🟡 Low — Settings label didn't match the danger level**

Updated to **"Dangerously skip all confirms by default"** with a description noting the blast radius ("The agent will run without asking for confirmation. Only honoured when the selected agent supports it.").

---

**Also — dead `setTaskStepsEnabled` re-introduced with zero callers**

Removed. It was deliberately dropped in `33b2e4f` and reintroduced by this PR.

johannesjo:
Thanks for the iterations here. The persistence wiring and dialog gating look mostly solid now, but I found one behavior I think should be fixed before merge.

Main issue: `createTask` still writes the per-task Steps checkbox choice back into `store.defaultStepsEnabled`:

- `src/store/tasks.ts`: `stepsEnabled !== store.defaultStepsEnabled` then `setStore('defaultStepsEnabled', stepsEnabled)`
- `src/components/NewTaskDialog.tsx`: the dialog always passes `stepsEnabled: stepsEnabled()`

That means checking or unchecking Steps for one task silently mutates the Settings -> New Task Defaults value. This preserves the old "last used `showSteps`" behavior, but the new UI presents this as an explicit default setting. The skip-permissions defaults do not mutate this way, so the behavior is inconsistent.

Suggested fix: only mutate `defaultStepsEnabled` from the Settings setter, not from task creation. It would be worth adding a regression test that creating a task with `stepsEnabled` opposite to `store.defaultStepsEnabled` does not change the default.

Smaller follow-ups:

- `src/store/persistence.ts` treats an invalid-but-present `defaultStepsEnabled` like it is absent and may fall back to legacy `showSteps`. If non-booleans should be rejected, the migration should distinguish "property absent" from "property present but invalid".
- `openspec/changes/new-task-defaults/tasks.md` still documents the old migration formula.
- `openspec/changes/new-task-defaults/change.md` mentions live task steps toggles updating the default, but I don't see a live toggle path.

Verification: `npm test -- src/store/persistence.test.ts src/store/autosave.test.ts src/components/NewTaskDialog.test.ts` passes with 71 tests.

brooksc:
All issues addressed in the latest push — in order:

---

**Main issue — `createTask` mutates `defaultStepsEnabled` on task creation**

Confirmed: `tasks.ts:249` had `if (stepsEnabled !== store.defaultStepsEnabled) setStore('defaultStepsEnabled', stepsEnabled)`, which silently wrote back the per-task checkbox choice into the app-level default on every task creation. The skip-permissions defaults had no such write-back, so the behavior was inconsistent.

Fix: removed lines 248–249 (the comment and the `setStore` call). `defaultStepsEnabled` is now only mutated by the Settings setter.

Regression test added to `src/store/tasks.test.ts`: creating a task with `stepsEnabled` opposite to `store.defaultStepsEnabled` asserts that `setStore` is never called with `'defaultStepsEnabled'`.

---

**Low — `persistence.ts` treats invalid-present `defaultStepsEnabled` like absent**

Confirmed: `typeof raw.defaultStepsEnabled === 'boolean' ? … : raw.showSteps === true` fell back to the legacy field even when `defaultStepsEnabled` was present but not a boolean (e.g. `"yes"`).

Fix: added a presence check so invalid present values resolve to `false` instead of falling back to `showSteps`:
```ts
s.defaultStepsEnabled =
  typeof raw.defaultStepsEnabled === 'boolean'
    ? raw.defaultStepsEnabled
    : 'defaultStepsEnabled' in (raw as object)
    ? false
    : raw.showSteps === true;
```

New persistence test added: `invalid present defaultStepsEnabled does not fall back to showSteps=true`.

---

**`openspec/changes/new-task-defaults/tasks.md` — outdated migration formula**

Updated the checklist line to describe the presence-check logic instead of the old `=== true || …` formula.

---

**`openspec/changes/new-task-defaults/change.md` — false live-toggle claim**

Removed the `WHEN the user creates a task or toggles steps on a live task → THEN defaultStepsEnabled is updated` clause. The spec now correctly states that the default is only updated via the Settings setter.

johannesjo:
Requested change before merge: the OpenSpec still says creating a task with the opposite Steps checkbox state updates `defaultStepsEnabled`, but the implementation now intentionally treats it as an explicit Settings-controlled default and the new test asserts task creation does not mutate it.

Please update both the canonical steps-tracking spec and the change spec to match the new contract:

- The New Task dialog reads `defaultStepsEnabled` when it opens.
- Settings toggles update `defaultStepsEnabled`.
- Creating a task with a different per-task checkbox value does not update the default.

I am not blocking on the missing component tests or the broader pre-existing autosave omissions.

brooksc:
## Spec update addressed

Thanks for the clear feedback. Both spec files have been updated in the latest push to match the actual contract.

---

### Canonical spec (`openspec/specs/steps-tracking/spec.md`)

**What was wrong:** The scenario titled "Default updates when user toggles the checkbox" described task creation as the trigger for mutating `defaultStepsEnabled`, which no longer matches the implementation.

**Fix:** Replaced that scenario with two new ones:
- **"Settings toggle updates the default"** — `defaultStepsEnabled` is mutated only when the user toggles the setting in Settings → General → New Task Defaults.
- **"Task creation does not update the default"** — creating a task with the checkbox in a different state leaves `defaultStepsEnabled` unchanged; only the per-task `stepsEnabled` flag reflects the checkbox state at creation time.

Also updated the stale `showSteps` reference to `defaultStepsEnabled` (the field was renamed in this PR).

---

### Change spec (`openspec/changes/new-task-defaults/specs/steps-tracking/spec.md`)

**What was wrong:** Same incorrect scenario ("Default updates when user toggles the checkbox") was present, asserting that task creation would update the flag.

**Fix:** Same replacement — the wrong scenario is removed and the two correct scenarios (Settings-only mutation, task-creation no-op) are in its place. The rest of the change spec (migration, skip-permissions/propagate additions, Settings configurability) is unchanged.

johannesjo:
Thank you very much! <3

PR #166, opened 2026-06-02
fix(worktree): prevent symlinked dirs from being committed via .git/info/exclude

brooksc:
## Problem

When a task is created with symlinked directories (e.g. `node_modules`),
git may treat the symlink as an untracked file and accidentally commit it.

**Root cause:** Many projects gitignore `node_modules` with a trailing slash
(`node_modules/`). Git's trailing-slash pattern only matches real
*directories* — a symlink named `node_modules` is a special file, not a
directory, so it doesn't match. When the agent runs `git add -A` during
normal work, the symlink is included and committed. On close/merge, the user
sees an unexpected outstanding commit containing `node_modules`.

**Why the `.gitignore` check doesn't catch it:** `getGitIgnoredDirs()` uses
`git check-ignore -q <name>` on the real directory in the main repo, which
correctly returns "ignored". But in the *worktree*, the same name refers to a
symlink, which the trailing-slash pattern doesn't match.

## Fix

After creating each symlink in `createWorktree()`, call the new
`ensureSymlinkExcludes()` function, which appends root-anchored patterns
without a trailing slash (e.g. `/node_modules`) to `.git/info/exclude`.

Root-anchored patterns without a trailing slash match **both symlinks and
directories**, so the exclusion holds regardless of how the project
`.gitignore` is written. This is the same mechanism already used by
`ensureSandboxExcludes()` for bwrap bind-mount artifacts.

The function is incremental: entries already present are not duplicated, and
new symlink names from later worktrees are added in subsequent calls (unlike
the all-at-once sandbox approach).

## Changes

**`electron/ipc/git.ts`**
- Added `SYMLINK_EXCLUDE_HEADER` constant
- Added `ensureSymlinkExcludes(worktreePath, symlinkNames)` — mirrors the
  structure of `ensureSandboxExcludes` but appends entries incrementally
- Updated `createWorktree()` to track which symlinks were successfully created
  and pass them to `ensureSymlinkExcludes`

## Tests

**`electron/ipc/git-symlink-excludes.test.ts`** — 14 new cases covering:

| Test | What it verifies |
|---|---|
| empty list | no-ops without touching git or fs |
| first call | writes header + root-anchored entry |
| no trailing slash | `/node_modules\n` not `/node_modules/\n` (symlink invariant) |
| correct file path | writes to `<commonDir>/info/exclude` |
| absolute commonDir | uses path as-is when git returns an absolute path |
| relative commonDir | joins against worktreePath |
| header already present | appends entry without duplicating header |
| name already present | skips entirely — no appendFileSync call |
| partial add | only appends the missing names |
| ENOENT on exclude file | treats as empty file, still writes |
| git rev-parse failure | silent no-op |
| newline prefix | adds `\n` when file doesn't end with one |
| no redundant prefix | no leading `\n` when file already ends with one |
| multiple names | single appendFileSync call for all new entries |

All 1118 tests pass (`npm test`).

### DIFF
--- a/electron/ipc/git-symlink-excludes.test.ts
+++ b/electron/ipc/git-symlink-excludes.test.ts
@@ -1,0 +1,195 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const { mockExecFileSync, mockReadFileSync, mockAppendFileSync } = vi.hoisted(() => ({
+  mockExecFileSync: vi.fn(),
+  mockReadFileSync: vi.fn(),
+  mockAppendFileSync: vi.fn(),
+}));
+
+vi.mock('child_process', async () => {
+  const actual = await vi.importActual<typeof import('child_process')>('child_process');
+  return {
+    ...actual,
+    execFileSync: mockExecFileSync,
+    execFile: vi.fn(),
+    spawn: vi.fn(() => ({
+      stdout: { on: vi.fn() },
+      stderr: { on: vi.fn() },
+      on: vi.fn(),
+    })),
+  };
+});
+
+vi.mock('fs', async () => {
+  const actual = await vi.importActual<typeof import('fs')>('fs');
+  return {
+    ...actual,
+    default: {
+      ...actual,
+      readFileSync: mockReadFileSync,
+      appendFileSync: mockAppendFileSync,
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much!

PR #167, opened 2026-06-02
fix(new-task-dialog): scroll to bottom when coordinator mode is enabled

### LINKED ISSUE #147: Adjust New Task window to deal with off-screen options
Below is an image of the New Task dialog and the Coordinator feature is enabled.
As you can see as you select a checkbox for Dangerously Skip Permissions - the warning text is off screen and requires scrolling.

More important there are now several options that are hidden from the user.  

I wanted to get input on how to address this, I see there are several options:
1. Increase the size of the window - height and perhaps width - to fit everything on screen even when options are selected.
2. When some text/options are "below the fold" then scroll the window down automatically so they are visible.
3. ...??  

Image

Image

brooksc:
Fixes #147.

## Problem

When the user checks **Coordinator mode** in the New Task Dialog, several
options expand below the fold — warning banner, max concurrent sub-tasks,
propagate skip-permissions, and the symlink section. There is no visual
indication that anything appeared; the user has to discover them by scrolling.

## Fix

Attach a ref to the scrollable container (`overflow-y: auto` div) and add a
`createEffect` that scrolls it to the bottom whenever `coordinatorMode` flips
to `true`:

```ts
createEffect(
  on(
    coordinatorMode,
    (enabled) => {
      if (enabled) {
        queueMicrotask(() => {
          if (scrollContainerRef) {
            scrollContainerRef.scrollTop = scrollContainerRef.scrollHeight;
          }
        });
      }
    },
    { defer: true },   // skip initial render — only fire on user toggles
  ),
);
```

Key details:
- **`defer: true`** — the effect is skipped on the dialog's first render so
  the dialog doesn't jump to the bottom when it opens (even if coordinator
  mode is on by default).
- **`queueMicrotask`** — defers the scroll until after Solid's DOM update so
  `scrollHeight` reflects the newly-rendered content.
- **Guard on `scrollContainerRef`** — safe against the early-mount edge case
  where the ref isn't yet assigned.

## Tests (`src/components/NewTaskDialog.test.ts`, 9 cases)

Following the project pattern of testing extracted logic functions:

| Test | What it verifies |
|---|---|
| scrolls to scrollHeight when enabled | core behaviour |
| no-op when disabled | doesn't scroll on uncheck |
| no-op when container is null | guard against missing ref |
| no-op when container is undefined | guard against early mount |
| scrolls to exact scrollHeight | no hardcoded offset |
| idempotent at bottom | safe to call repeatedly |
| scrollTop unchanged before microtask | timing contract |
| no scroll after microtask when false | disabled path via microtask |
| no error when container absent at microtask time | async guard |

All 1113 tests pass (`npm test`).

### DIFF
--- a/src/components/NewTaskDialog.test.ts
+++ b/src/components/NewTaskDialog.test.ts
@@ -1,12 +1,5 @@
 import { describe, expect, it } from 'vitest';
-
-// ---------------------------------------------------------------------------
-// scrollCoordinatorIntoView
-//
-// Pure helper mirroring the logic inside the createEffect that runs when
-// coordinator mode is toggled on. Tests verify: scrolls to scrollHeight when
-// enabled, no-ops when disabled, no-ops when the container is absent.
-// ---------------------------------------------------------------------------
+import { scrollCoordinatorIntoView } from './scrollCoordinatorIntoView';
 
 interface FakeScrollContainer {
   scrollTop: number;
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

PR #171, opened 2026-06-02
fix(autosend): check promptAppearedInOutput return value; retry on failed echo

brooksc:
## Problem

When the Codex auto-send fired at the wrong moment (e.g. mid-startup screen-redraw, partially addressed in #169), the prompt was written to the PTY but Codex discarded it. The `promptAppearedInOutput` helper correctly detected the failure and returned `false` — but that return value was **ignored**:

```typescript
// before — result discarded
await promptAppearedInOutput(props.agentId, val, preSendTail, signal);
```

The code then called `setAutoSentInitialPrompt(initialPromptSnapshot)` regardless, permanently blocking any retry. The user saw Codex ready at the prompt with no activity and their initial prompt gone.

## Fix

Check the return value. On failure, increment `autoSendRetry` — a new SolidJS signal that the auto-send `createEffect` tracks. Incrementing it re-runs the effect with fresh state (new stability checks, new agent-ready callback), so the prompt is re-sent once Codex is truly ready.

```typescript
const appeared = await pollUntilPromptAppearsInOutput(...);
if (!appeared) {
  if (untrack(autoSendRetry) < AUTO_SEND_MAX_RETRIES) {
    setAutoSendRetry((n) => n + 1);
  }
  return; // don't mark as delivered
}
```

Key details:
- **`initialPrompt` is not cleared** on failure — the `createEffect` restart picks up the same prompt
- **`AUTO_SEND_MAX_RETRIES = 2`** caps at 3 total attempts to prevent infinite loops
- After all retries fail the text stays in the field so the user can send manually
- The `finally { setSending(false) }` block still runs on early return ✓

## Implementation

The polling logic is extracted from a component-internal closure to a standalone exported module (`prompt-verify.ts`) with an injected `getTail` dependency. This makes it fully unit-testable without mounting the component.

## Tests (`prompt-verify.test.ts`, 12 cases)

| Test | What it verifies |
|---|---|
| empty prompt | returns true immediately, no polling |
| snippet in preSendTail | returns true immediately (pre-existing content) |
| ANSI in preSendTail | strips codes before comparing |
| first poll hit | returns true |
| after several polls | returns true after N misses |
| ANSI in getTail | strips codes from live tail |
| 40-char snippet | only first 40 chars of prompt used |
| timeout | returns false after deadline |
| multiple polls before timeout | verifies polling actually happens |
| pre-aborted signal | returns false immediately |
| signal aborted mid-poll | returns false |
| agentId pass-through | correct agentId forwarded to getTail |

All 1338 tests pass.

### DIFF
--- a/src/components/PromptInput.tsx
+++ b/src/components/PromptInput.tsx
@@ -1,9 +1,11 @@
-import { createSignal, createEffect, on, Show, onMount, onCleanup, untrack } from 'solid-js';
+import { createSignal, createEffect, on, Show, onMount, onCleanup, untrack, batch } from 'solid-js';
 import { fireAndForget, invoke } from '../lib/ipc';
 import { IPC } from '../../electron/ipc/channels';
 import {
   store,
   sendPrompt,
+  setInitialPrompt,
+  clearInitialPrompt,
   registerFocusFn,
   unregisterFocusFn,
   registerAction,
@@ -32,6 +34,7 @@ import { clearStagedNotification, setTaskTerminalInputPendingFromQuestion } from
 import { isLandedTaskState } from '../store/landing';
 import { processAutoFireTick } from './autofire-tick';
 import {
+  resolveAutoSendVerifyOutcome,
   shouldAckInitialPromptDelivery,
   shouldHandoffCoordinatorQuestion,
   shouldRendererAutoSendInitialPrompt,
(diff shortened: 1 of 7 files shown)


johannesjo (changes requested):
Requesting changes for one blocking lifecycle issue.

`src/components/PromptInput.tsx:779` in the PR retry branch does not actually retry the original initial prompt. `sendPrompt` treats this send as the queued initial prompt and clears `task.initialPrompt` once the write completes (`src/store/tasks.ts:645-675`). `TaskPanel` passes that same field into `PromptInput` as `initialPrompt`, and the autosend effect recomputes `ip` from `props.initialPrompt`. After `setAutoSendRetry(...)`, the effect reruns with `props.initialPrompt` already undefined and returns from `shouldRendererAutoSendInitialPrompt`, so no second attempt is scheduled.

Please keep a retry-owned snapshot of the prompt, or defer/restore clearing `initialPrompt` until delivery verification succeeds.

Secondary note: the extracted verifier has a few abort/deadline edge cases worth tightening (`signal.aborted` is checked after the empty/pre-send success paths, aborts during sleep wait until the next poll, and there is no final tail check at the deadline), but the store/lifecycle mismatch above is the blocker.

johannesjo:
Thank you very much! <3

Issue #151, opened 2026-05-27
Progress: "Completed today" counts closures instead of merges; metrics use inconsistent time dimensions


FourWindff:
---
  Progress Panel: "Completed today" counts closures instead of merges;
  metrics use inconsistent time dimensions

  Bug: "Completed today" increments on any close, not just completed work

  Reproduction steps:
  1. Create a new task, make no changes, and click Close.
  2. The "Completed today" counter in the sidebar increments from 0
  to 1.

  Root cause:

  recordTaskCompleted() is invoked unconditionally inside
  removeTaskFromStore() (src/store/tasks.ts:504). removeTaskFromStore
  is reached from multiple paths, including:

  - Normal close (closeTask) — covers empty tasks, unmerged tasks, etc. - Coordinator subtask closure (MCP_TaskClosed)

  Expected: "Completed today" should only count tasks that were successfully merged into main/master today.

  ---
  Design issue: The two metrics use different time dimensions

  The current Progress panel shows:
  - Completed today — resets daily
  - Merged to main/master — global cumulative total

  Problems:

  1. Semantic mismatch: Presented side-by-side, both numbers appear to
  be "today's" metrics, but Merged is actually a lifetime total.
  2. Readability decay: Over time, numbers like +15,000 / -3,200 become
   unwieldy and lose meaning as a daily progress indicator.
  3. Unclear purpose: Is Progress meant to show "what I did today" or
  "my total historical output"?

  ---
  Proposal: Unify Progress as a "Today's Work" panel

  Fix Completed today:
  - Rename to Merged today and only count tasks successfully merged
  today.
  - Increment inside mergeTask when cleanup=true; remove the
  unconditional call from removeTaskFromStore.

  Update Merged to main/master:
  - Switch to a daily tally to match the time dimension of Merged
  today.
  - Show today's +added / -removed.

  ---
  Future enhancement: Activity calendar (Contribution Graph)

  The value of Progress is helping users feel their daily rhythm. The
  current two-card layout is low-density. A small contribution
  calendar—similar to GitHub's profile graph or Claude Code Stats—would
   make daily activity much more tangible.

PR #178, opened 2026-06-13
fix(progress): gate merged-lines counter on cleanup

ASRagab:
## Summary

- `mergeTask` was calling `recordMergedLines()` unconditionally before the `if (cleanup)` gate, while `recordTaskMerged()` (the "Merged today" counter) was correctly gated inside the block — leftover asymmetry from the issue #151 fix.
- This let the two side-by-side sidebar stats ("Merged today" / "Merged (total)") drift apart on the non-cleanup merge path.
- Move `recordMergedLines(...)` inside `if (cleanup) { ... }` so both stats share one gate and the same source-of-truth event.

## Reproduce on `main`

1. Merge a task with the "merge & keep" option (`cleanup: false`).
2. `mergedLinesAdded` / `mergedLinesRemoved` bump; `completedTaskCount` does not.
3. Close the same task later via `closeTask` — "Merged today" is never incremented, but the lines totals already moved.
4. Sidebar footer shows lines added/removed for a task that "Merged today" never counted.

After this change both counters react to the same event (`cleanup === true`), matching the invariant locked in by the issue #151 fix ("only count tasks that complete the full merge+cleanup lifecycle").

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npx vitest run` — 1412 pass / 0 fail
- [x] Existing `recordTaskMerged counts merges with cleanup, not closures` describe extended with `recordMergedLines` parity assertions (called once with the merge result on `cleanup: true`; never called on `cleanup: false`).

### DIFF
--- a/src/store/tasks.test.ts
+++ b/src/store/tasks.test.ts
@@ -152,7 +152,7 @@ import {
   updateTaskBranch,
 } from './tasks';
 import { getCoordinatorChildren } from './sidebar-order';
-import { recordTaskMerged } from './completion';
+import { recordMergedLines, recordTaskMerged } from './completion';
 import { markAgentSpawned, rescheduleTaskStatusPolling } from './taskStatus';
 import { saveState } from './persistence';
 import { getProjectBranchPrefix, getProjectPath, isProjectMissing } from './projects';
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #180, opened 2026-06-16
fix(agent): fallback to fresh start when resume args fail

FourWindff:
## Summary

When using Claude Code (which supports session resumption via `--continue` ) to create a task, if the user closes and restarts the app before any conversation has taken place, the system relaunches the agent using the `resume_args` defined in `AgentDef`. Because no session exists to continue, the process exits immediately with `No conversation found to continue`, forcing the user to manually click the restart button to recover the task. 

## 1. Root Cause

- After the app restarts, agents in the `resumed` state are launched with `AgentDef.resume_args` (for Claude Code this is typically `--continue` ).
- If no conversation took place before the restart, Claude Code cannot find a session to resume and exits with an error.
- The previous implementation had no automatic fallback when an agent exited due to failed resume arguments, leaving the user to manually restart.

## 2. Solution

This PR adds resume-failure detection in the `onExit` callback of `TaskAITerminal`:

- If the agent is in `resumed` state, exits with a non-zero code, and its final output matches a known resume-failure signature (e.g. Claude's `No conversation found to continue`), it automatically calls `restartAgent(id, false)`.
- Passing `false` skips `resume_args` and falls back to `AgentDef.args`, starting a fresh session without requiring manual intervention.

It also introduces `isResumeArgsFailure` in `src/lib/agent-args.ts` to centralize the resume-failure signatures for each CLI, making future extensions easier.

### DIFF
--- a/src/components/TaskAITerminal.tsx
+++ b/src/components/TaskAITerminal.tsx
@@ -30,7 +30,7 @@ import { createHighlightedMarkdown } from '../lib/marked-shiki';
 import type { Task } from '../store/types';
 import type { AgentDef } from '../ipc/types';
 import type { PromptInputHandle } from './PromptInput';
-import { buildTaskAgentArgs } from '../lib/agent-args';
+import { buildTaskAgentArgs, isResumeArgsFailure } from '../lib/agent-args';
 
 function aiTerminalPanelId(agentId: string): string {
   return `ai-terminal:${agentId}`;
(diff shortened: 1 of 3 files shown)


johannesjo:
## Review

Nice, focused fix — I traced the full flow and the core logic is correct and, importantly, **loop-safe**: `restartAgent(id, false)` sets `resumed = false`, so if the fresh start also exits non-zero the `a().resumed` guard is false and it falls through to `markAgentExited` rather than restarting again. Going straight from running → running (skipping the "exited" state) is the right call for a seamless recovery, and centralizing the signatures in `RESUME_FAILURE_PATTERNS` is a clean, extensible approach.

A few things worth addressing:

**Important**
- **Add a unit test for `isResumeArgsFailure`.** It's a pure function and `src/lib/agent-args.test.ts` already exists right beside it. Suggested cases: claude match, claude no-match, non-claude command (→ false), empty `lastOutput` (→ false), and a multi-line `lastOutput` where the pattern is on one of several lines.

**Minor**
- **Manual "Resume" button now silently falls back too.** The new `onExit` logic also fires when the user explicitly clicks **Resume** and it fails — the agent restarts fresh with no indication the resume failed. For the app-restart auto-resume case that silence is fine; just worth a conscious decision on whether an *explicit* Resume click should surface a brief notice.
- **English-string matching is inherently fragile.** `No conversation found to continue` will break if Claude rewrites/localizes the message, and the raw `last_output` carries ANSI escapes (only `\r` is stripped in `pty.ts`). The substring match still works as long as the codes bracket the phrase rather than split it — low risk for a short non-wrapping string — but it's a signature that'll need maintenance. Centralizing it makes that easy, so this is just a flag.
- **`code.exit_code !== 0` also matches `null`** (signal kill). Harmless because `isResumeArgsFailure` must also match, but noting it.

**Verdict:** Ready to merge once the `isResumeArgsFailure` test is added.

johannesjo:
Thank you very much!

Issue #122, opened 2026-05-16
Desktop Notifications Never Appear on macOS


brooksc:
**App:** Parallel Code v1.9.0  
**Platform:** macOS 26.4.1 (Build 25E253)  
**Electron:** v40.8.5  
**Node:** v25.9.0  

---

### Summary

Desktop notifications are enabled in Settings but never fire on macOS. The app does not appear in **System Settings > Notifications**, which is required for macOS to deliver notifications.

### Steps to Reproduce

1. Launch Parallel Code on macOS
2. Open Settings and enable "Desktop notifications"
3. Run a task to completion
4. Observe: no notification appears
5. Open **System Settings > Notifications** — Parallel Code is not listed

### Expected Behavior

On first enabling desktop notifications, macOS should prompt the user to allow notifications. After granting permission, the app should appear in System Settings > Notifications and deliver notifications when tasks complete or need attention.

### Actual Behavior

No permission prompt is ever shown. No notifications are delivered. The app is absent from System Settings > Notifications.

### Root Cause

`systemPreferences.requestNotificationPermission()` is never called. macOS 10.14+ requires explicit permission before an app can deliver notifications — without this call, the OS silently drops all notification attempts.

The code in `electron/ipc/register.ts:781` checks `Notification.isSupported()`, which returns `true` regardless of permission status, masking the failure. The notification objects are created and `.show()` is called, but macOS discards them silently.

**Relevant code paths:**
- `electron/ipc/register.ts:779–812` — notification IPC handler (missing permission request)
- `electron/ipc/pr-checks.ts:275–302` — PR check notifications (same issue)
- `src/store/desktopNotifications.ts:22` — checks `desktopNotificationsEnabled` flag but no permission gate

### Fix

Call `systemPreferences.requestNotificationPermission()` in the Electron main process:
- When the user enables desktop notifications (via a new IPC call from the renderer)
- On app startup if `desktopNotificationsEnabled` is already `true` (to fix existing users)

The method returns a promise resolving to `'granted'`, `'denied'`, or `'default'`, which should be surfaced back to the renderer so the UI can reflect the actual permission state.

---

I'll take a pass at fixing and submitting later

brooksc:
odd - I'm running from a branch w/ npm run dev.  and notifications are coming through fine on this.  just not on the released version.  will investigate later.

PR #185, opened 2026-06-22
fix: surface notification failure to renderer on macOS

xbrxr03:
## Summary

Surface notification `failed` events to the renderer so the UI can inform the user when a notification did not deliver.

On macOS, unsigned development builds silently drop notifications because `UNNotification` requires code-signing. The `failed` event currently fires only on Windows (Electron 40). Starting with Electron 42, it will fire on macOS too — this PR lays the groundwork by wiring the IPC channel now.

Refs #122

## Changes

- Add `IPC.NotificationFailed` channel + preload allowlist entry
- In `register.ts`, forward the `failed` event to the renderer via `NotificationFailed` (includes error string + platform)
- In `desktopNotifications.ts`, add a listener that logs a macOS-specific `console.warn` explaining unsigned-build requirements
- Update the watcher-replacement test to cover both `NotificationClicked` and `NotificationFailed` listener cleanup

## Reviewer notes

- `taskIds` removed from the `NotificationFailed` payload — the renderer listener does not use them
- Comments updated to note the Electron 42+ scope for macOS
- The `console.warn` in the renderer is a DevTools-visible diagnostic. A proper toast/banner can follow when the settings-level alerts design lands

### DIFF
--- a/electron/ipc/channels.ts
+++ b/electron/ipc/channels.ts
@@ -133,6 +133,7 @@ export enum IPC {
   // Notifications
   ShowNotification = 'show_notification',
   NotificationClicked = 'notification_clicked',
+  NotificationFailed = 'notification_failed',
 
   // PR CI status
   StartPrChecksWatcher = 'start_pr_checks_watcher',
--- a/electron/ipc/register.ts
+++ b/electron/ipc/register.ts
@@ -1040,11 +1040,12 @@ export function registerAllHandlers(win: BrowserWindow): void {
         release();
         logWarn('notification', 'show failed', { ...warnCtx, err: error });
         // Surface the failure to the renderer so the UI can inform the user.
-        // On macOS, unsigned development builds fail silently because
-        // UNNotification requires code-signing. This lets the user know.
+        // NOTE: The `failed` event only fires on macOS starting in Electron 42
+        // (UNNotification migration). On Electron 40 (current), this code path
+        // is macOS-inert — it still fires on Windows. Once the project upgrades
+        // past Electron 42, this will activate for unsigned macOS builds too.
         if (!win.isDestroyed()) {
           win.webContents.send(IPC.NotificationFailed, {
-            taskIds: args.taskIds,
             error: typeof error === 'string' ? error : String(error),
             platform: process.platform,
           });
--- a/electron/ipc/register.ts
+++ b/electron/ipc/register.ts
@@ -1039,6 +1039,16 @@ export function registerAllHandlers(win: BrowserWindow): void {
       notification.on('failed', (_event, error) => {
         release();
         logWarn('notification', 'show failed', { ...warnCtx, err: error });
+        // Surface the failure to the renderer so the UI can inform the user.
+        // On macOS, unsigned development builds fail silently because
+        // UNNotification requires code-signing. This lets the user know.
+        if (!win.isDestroyed()) {
+          win.webContents.send(IPC.NotificationFailed, {
+            taskIds: args.taskIds,
+            error: typeof error === 'string' ? error : String(error),
+            platform: process.platform,
+          });
+        }
       });
       notification.on('click', () => {
         release();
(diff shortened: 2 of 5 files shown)


johannesjo:
Thanks for digging into this — the diagnosis is sharp. You correctly spotted that the `requestNotificationPermission()` fix proposed in #122 is obsolete and that code-signing is the real constraint on modern Electron. The diff is clean, the `win.isDestroyed()` guard and `String(error)` coercion are right, and the test update properly models the new two-listeners-per-watcher ordering.

One blocking concern before this can land as a fix for #122:

### The `failed`-on-macOS behavior only exists in Electron 42+, but this repo is on Electron 40

The whole mechanism here depends on `notification.on('failed')` firing for unsigned macOS builds. That behavior — and the code-signing requirement — was introduced by the `NSUserNotification` → `UNUserNotificationCenter` migration that shipped in **Electron 42.0**:

- Breaking change *"macOS notifications now use `UNNotification` API (42.0)"*: "Electron has migrated from the deprecated `NSUserNotification` API to the `UNNotification` API on macOS. The new API requires that an application be code-signed in order for notifications to be displayed." ([breaking-changes](https://www.electronjs.org/docs/latest/breaking-changes), [electron#47817](https://github.com/electron/electron/pull/47817))
- In the **v40.8.5** docs the `failed` event is annotated **`_Windows_` only** — it does not fire on macOS, and there's no code-signing note. (Compare with `main`, where it's now `_macOS_ _Windows_`.)

This repo pins `electron: "^40.8.5"` (resolved install: `40.8.5`), and `^40` can't float to 42. So on the **currently shipped build the new `console.warn` will essentially never fire on macOS** — issue #122 users get the same silence as before. The PR description itself says "Electron 42+", which is the tell: the change is correct for a version we're not on yet.

### Knock-on points

- **`Closes #122` overstates it.** #122 asks for notifications to actually work (prompt → appear in System Settings → deliver). This is a diagnostic only, and an inert one on Electron 40. Suggest `Refs #122` and leaving the issue open. Worth noting a collaborator reported notifications working on a `npm run dev` branch but *not* the released build — the inverse of what code-signing alone predicts, which hints the released-app signing/entitlement is the real gap to chase.
- **The user is never actually informed.** Both code comments say "so the UI can inform the user", but the renderer only does `console.warn`, which lives in DevTools and is invisible to someone who just toggled a setting. Even on Electron 42 this wouldn't meet the stated intent without a real surface (settings banner / toast) driven off the event.
- Minor: `taskIds` is forwarded on failure but unused by the renderer listener.

### Suggested path

Any of: (a) hold this until the Electron 42 upgrade and bundle it there; (b) keep it but pair it with a user-facing surface instead of a DevTools log; or (c) merge as deliberately forward-looking groundwork — with `Closes #122` downgraded and a comment noting it activates on Electron 42+.

*Sources: [Electron 42 release notes](https://www.electronjs.org/blog/electron-42-0) · [electron/electron#47817](https://github.com/electron/electron/pull/47817) · [Notification API docs](https://www.electronjs.org/docs/latest/api/notification) · [Breaking Changes](https://www.electronjs.org/docs/latest/breaking-changes)*

xbrxr03:
Thanks for the thorough review! I've pushed an update addressing all three points:

1. **`Closes #122` → `Refs #122`** — Updated the PR description and scope. This is forward-looking groundwork that activates after an Electron 42 upgrade, not a full fix for the current version.

2. **Removed `taskIds` from the payload** — The renderer listener doesn't use them, so they're gone. The `NotificationFailed` event now sends only `error` and `platform`.

3. **Updated all comments** to explicitly note the Electron 42+ scope for macOS — both in `register.ts` and `desktopNotifications.ts`. The renderer handler logs a DevTools-visible `console.warn` for macOS. A proper toast/banner can land separately when the settings-level alerts design is ready.

All 1420 tests pass, typecheck + lint clean.

johannesjo:
Thank you very much ! <3

Issue #157, opened 2026-05-30
[Quality audit] Fix correctness bug and red repository gates


johannesjo:
Derived from the local audit report `CODE_QUALITY_FINDINGS.md` created on 2026-05-30.

## Scope

This issue tracks the highest-priority correctness and quality-gate failures found during the audit.

## Findings

- [ ] Finding 1: `wait_for_signal_done` timeout bypasses replay cache.
  - Evidence: `electron/mcp/coordinator.ts:1626-1708`, especially timeout path around `1685-1699`; `electron/mcp/replay-cache.ts`.
  - Fix direction: use one `complete(result)` path for timer cleanup, replay-cache write, resolver removal, active-wait handling, and promise resolution.
  - Tests: add retry-after-timeout coverage for the same `requestId`.
- [ ] Finding 2: store import cycles break `lint:arch`.
  - Evidence: `npm run lint:arch` reports 7 cycles across `src/store/projects.ts`, `tasks.ts`, `focus.ts`, `navigation.ts`, `persistence.ts`, and `terminals.ts`.
  - Fix direction: extract pure selectors/helpers so store modules flow one way.
- [ ] Finding 3: CI skips major repo gates.
  - Evidence: `.github/workflows/ci.yml` runs typecheck/lint/format only; Electron code is mostly compiled by `npm run compile`.
  - Fix direction: run `npm run check` and `npm test` in CI; add `npm run check:static` after current failures are fixed.
- [ ] Finding 7: Knip reports verified dead files, exports, types, dependency, and config drift.
  - Evidence: `npm run lint:dead` reports stale entry points, unused files, unused exports/types, unused `@types/dompurify`, and unlisted binaries.
  - Fix direction: update Knip entries, remove verified dead code/dependency, and keep the gate green.
- [ ] Finding 38: OpenSpec validation is red and completed changes remain active.
  - Evidence: `npx --no-install openspec spec validate custom-themes --strict` fails with `requirements.4.text: Requirement must contain SHALL or MUST keyword`; `npx --no-install openspec change validate custom-themes --strict` fails because the change has no delta specs.
  - Fix direction: fix custom-theme spec structure, add/remove the dangling change, and archive completed changes.
- [ ] Finding 41: npm config and security-rule script are noisy/brittle.
  - Evidence: `.npmrc` emits npm warnings; `npm run test:security-rules` fails with `spawnSync semgrep ENOENT` when semgrep is missing.
[...]

PR #186, opened 2026-06-22
fix: replace noisy console.warn with structured logger in steps.ts

xbrxr03:
The step watcher was logging every send and every file change at `console.warn` level — that's debug info, not a warning. In a normal session with multiple agents running, this spams the console with hundreds of entries that make real warnings impossible to spot.

Switched all logging in `steps.ts` to the structured logger that the rest of the main process already uses:
- Routine events (step sends, watch events) → `info` level
- Actual failures (write errors, watcher errors) → `warn` level
- All entries use structured context objects instead of string concatenation

Found during quality audit (#157, finding 41).

### DIFF
--- a/electron/ipc/steps.ts
+++ b/electron/ipc/steps.ts
@@ -2,6 +2,7 @@ import fs from 'fs';
 import path from 'path';
 import type { BrowserWindow } from 'electron';
 import { IPC } from './channels.js';
+import { info as logInfo, warn as logWarn, errMessage } from '../log.js';
 
 interface StepsWatcher {
   fsWatcher: fs.FSWatcher | null;
@@ -29,7 +30,7 @@ const processedCount = new Map<string, number>();
 function sendStepsContent(win: BrowserWindow, taskId: string, stepsFile: string): void {
   if (win.isDestroyed()) return;
   const steps = readStepsFile(stepsFile);
-  console.warn('[steps.send]', taskId, 'len=', steps?.length ?? 'null');
+  logInfo('steps', 'send', { taskId, len: steps?.length ?? null });
   if (steps) applyTimestamps(steps, stepsFile, taskId);
   win.webContents.send(IPC.StepsContent, { taskId, steps });
 }
@@ -64,7 +65,7 @@ function applyTimestamps(steps: unknown[], stepsFile: string, taskId: string): v
   try {
     fs.writeFileSync(stepsFile, JSON.stringify(steps, null, 2), 'utf-8');
   } catch (err) {
-    console.warn('[steps] Failed to write back timestamps:', err);
+    logWarn('steps', 'failed to write back timestamps', { err: errMessage(err) });
   }
 }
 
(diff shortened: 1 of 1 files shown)


johannesjo:
## Review — structured logger migration in `steps.ts`

Clean, well-targeted change. The level mapping is the right fix: `info` is suppressed under the prod `warn` floor, so the console spam disappears in real sessions while staying visible in dev, and the structured context objects are a clear improvement over string concatenation. Verified the logger API (`info`/`warn`/`errMessage`) and the `../log.js` ESM import path against the call sites — all correct.

One gap and two minor nits:

**Important — one `console.error` is left behind (contradicts "Switched all logging").**
`readStepsFile` at `electron/ipc/steps.ts:80` still does:
```ts
console.error('[steps] Failed to read steps file:', e);
```
This is the one genuine *error* path in the file (corrupt/unreadable `steps.json`, non-ENOENT), and it's exactly what `log.ts`'s `error(category, msg, err, ctx)` exists for — it also captures the stack. Suggest:
```ts
import { error as logError, info as logInfo, warn as logWarn, errMessage } from '../log.js';
// ...
logError('steps', 'failed to read steps file', e);
```

**Minor — `watch event` logs before the filename filter.**
At `steps.ts:160`, the `logInfo('steps', 'watch event', …)` fires for *every* file change in `.claude/`, not just `steps.json` (the `filename \!== 'steps.json'` early-return is on the next line). It's the highest-frequency log here, so `debug` reads as more honest than `info` — relevant if a user lowers the floor to `info` via the renderer's `level_min`, which would re-introduce spam one level down. No-op under the current prod/dev floors, so purely a level-choice nuance.

**Minor — stack traces dropped for watch-setup failures.**
`errMessage(err)` keeps only `err.message`. For `failed to watch worktree root` / `failed to watch steps dir` (`steps.ts:194,220`), the failure silently disables live step updates for that task. The old `console.warn(msg, err)` printed the full Error incl. stack; consider `error` level here so the logger captures the stack for the cases that actually break the feature. Fine to keep as `warn` if these are treated as recoverable.

Otherwise good to merge.

<sub>🤖 Automated review via Claude Code</sub>

xbrxr03:
Thanks for the thorough review! Pushed a follow-up commit addressing all three points:

1. **Remaining \`console.error\`** — replaced with \`logError' in \`readStepsFile\`. It now captures the full stack trace through the structured logger instead of just printing to stderr.

2. **Watch-event \`info\` → \`debug\`** — the \`onChange\` callback now logs at \`debug\` level, so it won't fire at all under the default \`warn\` floor. This is the highest-frequency log site in the file, so the level choice matters.

3. **Watcher setup errors → \`error\` level** — the three watcher error handlers (\`parentWatcher.on('error', …)\`, the \`fs.watch\` catch, and \`attachStepsDirWatcher\`'s catch/error) now use \`logError\` instead of \`logWarn\`. These failures silently disable live step updates, so the stack trace capture matters. The \`errMessage()\` wrapper is gone — \`logError\` already serialises and clips stacks internally.

All 1420 tests pass, typecheck + lint clean.

johannesjo:
Thank you very much! <3

PR #187, opened 2026-06-22
fix: clean up landedOrderCounters on coordinator close

xbrxr03:
The `landedOrderCounters` Map accumulates entries keyed by `coordinatorTaskId` but never deletes them when a coordinator closes. Over a long session with many coordinator cycles, these stale entries grow without bound — a slow memory leak.

Every other Map in the same cleanup path (`coordinators`, `activeSignalWaitCounts`, etc.) is already cleaned up on coordinator close. `landedOrderCounters` was just missed.

One line: add `this.landedOrderCounters.delete(coordinatorTaskId)` next to the existing `this.coordinators.delete(coordinatorTaskId)` call.

### DIFF
--- a/electron/mcp/coordinator.ts
+++ b/electron/mcp/coordinator.ts
@@ -2216,6 +2216,7 @@ export class Coordinator {
     }
 
     this.coordinators.delete(coordinatorTaskId);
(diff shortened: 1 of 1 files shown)


johannesjo:
Thank you very much! <3

PR #188, opened 2026-06-28
fix(copilot): use --additional-mcp-config so Copilot does not exit before prompt

### LINKED ISSUE #146: Copilot CLI Agent Exits Before Prompt Sent
Copilot CLI is listed as a supported agent in parallel-code (README, agents.ts), but it doesn't work when spawned as a subprocess via the `-p "{prompt}"` command-line invocation. The process exits before the prompt is delivered or processed.

## Reproduction Steps

1. Open parallel-code UI
2. Select Copilot CLI as an agent
3. Enter a task prompt
4. Attempt to create a task
5. Observe: Agent fails with "Agent exited before prompt was sent"

anxkhn:
Fixes #146

## Problem

Parallel Code runs agents as interactive PTY REPLs and types the queued prompt once the TUI is ready. With Copilot, the process exits on launch before the prompt is sent, surfacing as "Agent exited before prompt was sent".

Root cause: when the MCP coordinator is wired in, Copilot was grouped with the Claude-compatible agents and launched with `--mcp-config <path>`. Copilot has no such flag and exits immediately:

    $ copilot --mcp-config /tmp/x.json
    error: unknown option '--mcp-config'

This happens whenever Copilot runs as a coordinator agent or a coordinated sub-agent (the core multi-agent workflow).

## Fix

Use Copilot's real flag. Copilot reads MCP config from `--additional-mcp-config <@file|json>` and auto-discovers the workspace `.mcp.json` that Parallel Code already writes; both take the same `{ "mcpServers": {...} }` shape. Map Copilot to `--additional-mcp-config @<configPath>` in both the backend launch-arg builder (`electron/mcp/agent-args.ts`) and the renderer fallback (`src/lib/agent-args.ts`), mirroring the existing Codex and Antigravity special cases. Also wire Copilot session resume via `--continue`.

## Testing (on this machine, copilot v1.0.65, real PTY)

- Before: `copilot --mcp-config @file` exits with "unknown option". After: `copilot --additional-mcp-config @file` launches and stays interactive.
- `copilot mcp list` confirms it auto-discovers the app's `.mcp.json` (`parallel-code (local)`).
- `copilot --continue` with no prior session starts fresh without error.
- `npm run typecheck`, `npm run lint` (eslint, max-warnings 0), `npm run format:check`, `npm run lint:arch` (dependency-cruiser), and the full `npm test` (vitest, 1433 tests) pass. Added unit tests for the Copilot command detection, the `--additional-mcp-config` mapping, and the resume arg ordering.

### DIFF
--- a/electron/ipc/agents.ts
+++ b/electron/ipc/agents.ts
@@ -58,7 +58,7 @@ const DEFAULT_AGENTS: AgentDef[] = [
     name: 'Copilot CLI',
     command: 'copilot',
     args: [],
-    resume_args: [],
+    resume_args: ['--continue'],
     skip_permissions_args: ['--yolo'],
     description: "GitHub's Copilot CLI agent",
     // Copilot CLI shows up to two init dialogs (folder trust + instructions init)
--- a/electron/mcp/agent-args.test.ts
+++ b/electron/mcp/agent-args.test.ts
@@ -5,6 +5,7 @@ import {
   buildMcpLaunchArgs,
   isAntigravityCommand,
   isCodexCommand,
+  isCopilotCommand,
 } from './agent-args.js';
 
 const config = {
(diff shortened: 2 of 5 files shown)


johannesjo:
Thank you very much! <3

PR #194, opened 2026-07-01
fix(diff): show deleted file hunks

FourWindff:
## Summary

- Show deleted file hunks in the diff viewer instead of replacing them with a single deleted-file banner.
- Align deleted file rendering with added and modified file diff blocks, so the changed code remains visible.
- Remove the now-unused deleted-file banner styling import.

## Why

Deleted files previously only showed a single "This file was deleted" message in the diff block. That made deleted-file diffs inconsistent with added and modified files, and it prevented users from seeing what an agent actually removed.

## Test Plan

- [x] npm run typecheck
- [x] git diff --check


johannesjo:
Thank you very much! <3

PR #196, opened 2026-07-03
Fix quality audit repository gates

ASRagab:
## Summary
- Clean up Knip configuration so the dead-code gate runs without stale config hints.
- Remove the last stale OpenSpec repo reference now that OpenSpec has been removed.
- Expand CI to run `check:static` and the Semgrep filesystem-safety regression test.

## Notes
- Addresses #157.
- Semgrep remains an external CLI and is installed in CI with `pipx install semgrep` rather than adding the placeholder npm package.
- Existing coordinator replay-cache and store import-cycle fixes were left untouched and verified through the existing tests/gates.

## Verification
- `npm run lint:arch`
- `npm run lint:dead`
- `export PATH="$HOME/.local/bin:$PATH"; npm run test:security-rules`
- `rg -n --hidden --glob '!node_modules/**' --glob '!.git/**' --glob '!docs/**' 'openspec|OpenSpec' .` exits 1 with no matches
- `npm run check`
- `npm test`
- `gh workflow view CI --repo johannesjo/parallel-code`

## Review
- Independent code review found no Critical or Important findings.
- Minor follow-up only: consider pinning Semgrep if CI drift becomes a concern.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,11 +22,16 @@ jobs:
       - name: Check (compile + typecheck + lint + format)
         run: npm run check
 
-      - name: Lint (architecture)
-        run: npm run lint:arch
+      - name: Check static (typecheck + lint + dead code + architecture)
+        run: npm run check:static
 
-      - name: Lint (dead code)
-        run: npm run lint:dead
+      - name: Install Semgrep
+        run: |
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

PR #215, opened 2026-07-07
Use shared Dialog for modal overlays

### LINKED ISSUE #209: MCPLogModal and Arena CommitDialog bypass Dialog.tsx (no Escape, no focus trap)
## Problem

Two modals bypass the shared `Dialog` component and lose all of its behavior — no Escape-to-close, no focus trap, no `role="dialog"`/`aria-modal`, no focus restore. They can only be dismissed by clicking the backdrop.

- `SubTaskStrip.tsx:31-56` (`MCPLogModal`): `position: fixed; inset: 0` div with backdrop `onClick={() => props.onClose()}`.
- `src/arena/CommitDialog.tsx:22-24`: identical pattern (only an `autofocus` on the input).

`Dialog.tsx` already provides everything (focus trap via `src/lib/focus-trap.ts`, focus restore, topmost-only Escape via the dialog stack, `aria-modal` — `:25-50,104-105`), and the trap/restore helpers are importable on their own.

## Proposed fix

Render both through `Dialog`, or at minimum add the Escape listener + `createFocusTrap`/`createFocusRestore`. Small diff, removes real keyboard traps — good first issue.

---

LarryHu0217:
## Summary
- render MCPLogModal through the shared Dialog component
- render Arena CommitDialog through Dialog while keeping existing content styling
- remove the now-unused arena commit overlay CSS

Fixes #209.

## Tests
- npm run check
- npm run test -- src/lib/dialog-stack.test.ts src/lib/dialog.test.ts
- npm run test

### DIFF
--- a/src/arena/CommitDialog.tsx
+++ b/src/arena/CommitDialog.tsx
@@ -1,4 +1,5 @@
-import { Show, createSignal, untrack } from 'solid-js';
+import { Show, createSignal, createUniqueId, untrack } from 'solid-js';
+import { Dialog } from '../components/Dialog';
 import { arenaStore } from './store';
 import type { BattleCompetitor } from './types';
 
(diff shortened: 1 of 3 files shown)


johannesjo:
Thank you very much! <3

PR #216, opened 2026-07-07
Gate close task confirmation while status loads

### LINKED ISSUE #199: Delete button in CloseTaskDialog is enabled before the dirty-worktree check resolves
## Problem

In `CloseTaskDialog`, the red **Delete** button is enabled before the dirty-worktree check resolves, so a task with uncommitted work can be permanently deleted before the warning ever renders.

- The uncommitted/unmerged check is an async `createResource` (`CloseTaskDialog.tsx:17-23`); the "Warning: There are uncommitted changes that will be permanently lost." banner renders only once it resolves (`:71-92`).
- The `ConfirmDialog` invocation passes no `confirmDisabled`/`confirmLoading`, so Delete is clickable immediately.
- `MergeDialog` already does this correctly: its confirm is gated on the worktree status having loaded (`MergeDialog.tsx:520-522`).

Mitigating factors (why this is a narrow race rather than a common footgun): `ConfirmDialog` auto-focuses **Cancel** (`ConfirmDialog.tsx:29-41`), so Enter can't confirm, and a local `GetWorktreeStatus` usually resolves fast. But on a large repo / slow disk the window is real, and the failure mode is unrecoverable data loss.

## Proposed fix

Pass `confirmDisabled={worktreeStatus.loading}` (or `confirmLoading`) to the ConfirmDialog, mirroring the MergeDialog pattern.

Related smaller improvement: whether the branch is deleted on close is controlled by a project-level setting rather than visible in the dialog — consider a "Delete branch" checkbox in the close dialog itself, pre-set from the project default (see also the label mismatch tracked in #214: the setting's label says "on merge" but it's read on close).

---

LarryHu0217:
## Summary
- disable the CloseTaskDialog confirm action while internal worktree status is still loading
- show the existing confirm loading spinner during that pending state
- add focused coverage for the confirm gating logic

Fixes #199.

## Why
CloseTaskDialog checks dirty/unmerged worktree state asynchronously, but the destructive Delete action was clickable before that status resolved. This created a small race where uncommitted work could be deleted before the warning rendered.

## Tests
- npm run test -- src/components/CloseTaskDialog.test.ts
- npm run typecheck
- npm run check
- npm run test

### DIFF
--- a/src/components/CloseTaskDialog.test.ts
+++ b/src/components/CloseTaskDialog.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, it } from 'vitest';
+import { shouldDisableCloseTaskConfirm } from './CloseTaskDialog';
+
+describe('shouldDisableCloseTaskConfirm', () => {
+  it('disables confirmation for internal worktree tasks while status is loading', () => {
+    expect(
+      shouldDisableCloseTaskConfirm({ gitIsolation: 'worktree', externalWorktree: false }, true),
+    ).toBe(true);
+  });
+
+  it('enables confirmation after internal worktree status loads', () => {
+    expect(
+      shouldDisableCloseTaskConfirm({ gitIsolation: 'worktree', externalWorktree: false }, false),
+    ).toBe(false);
+  });
+
+  it('does not block imported external worktrees', () => {
+    expect(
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much!

PR #217, opened 2026-07-07
Confirm missing-folder project removal

### LINKED ISSUE #200: 'Remove' in EditProjectDialog missing-folder banner closes all project tasks without confirmation
## Problem

The **Remove** button in `EditProjectDialog`'s missing-folder banner closes every task in the project with a single click and no confirmation:

```
EditProjectDialog.tsx:200-205
  → await removeProjectWithTasks(project().id); props.onClose();
```

`removeProjectWithTasks` (`src/store/project-cleanup.ts:10-30`) sequentially closes all of the project's tasks (coordinators last).

The sidebar's remove-project action for the *same operation* goes through a ConfirmDialog (`Sidebar.tsx:585-587`) — so this is an inconsistency, not a policy.

A missing project folder (unmounted volume, renamed directory) is exactly the situation where users click around exploring, and worktrees/branches may still exist even though the project root is gone.

## Proposed fix

Route this button through the same ConfirmDialog as the sidebar path, stating how many tasks will be closed.

---

LarryHu0217:
## Summary
- route the missing-folder Remove action in EditProjectDialog through ConfirmDialog
- show the same task-count-aware project removal copy used by the Sidebar path
- share project removal confirmation helpers between both call sites
- add focused coverage for task counting and confirmation copy

Fixes #200.

## Why
The missing-folder banner removed a project and closed its tasks immediately after one click. That is the same destructive operation as Sidebar project removal, so it should use the same confirmation flow before closing tasks and deleting their worktrees/branches.

## Tests
- npm run test -- src/components/project-remove-confirmation.test.ts
- npm run typecheck
- npm run check
- npm run test

### DIFF
--- a/src/components/EditProjectDialog.tsx
+++ b/src/components/EditProjectDialog.tsx
@@ -1,6 +1,7 @@
 import { createSignal, createEffect, For, Show } from 'solid-js';
 import { Dialog } from './Dialog';
 import {
+  store,
   updateProject,
   PASTEL_HUES,
   isProjectMissing,
@@ -12,6 +13,12 @@ import { theme, sectionLabelStyle } from '../lib/theme';
 import type { Project, TerminalBookmark, GitIsolationMode } from '../store/types';
 import { SegmentedButtons } from './SegmentedButtons';
 import { ImportWorktreesDialog } from './ImportWorktreesDialog';
+import { ConfirmDialog } from './ConfirmDialog';
+import {
+  getProjectTaskCount,
+  projectRemoveConfirmLabel,
+  projectRemoveConfirmMessage,
+} from './project-remove-confirmation';
 
 interface EditProjectDialogProps {
   project: Project | null;
@@ -34,12 +41,17 @@ export function EditProjectDialog(props: EditProjectDialogProps) {
   const [bookmarks, setBookmarks] = createSignal<TerminalBookmark[]>([]);
   const [newCommand, setNewCommand] = createSignal('');
   const [showImportDialog, setShowImportDialog] = createSignal(false);
+  const [confirmRemoveProjectId, setConfirmRemoveProjectId] = createSignal<string | null>(null);
   let nameRef!: HTMLInputElement;
 
   // Sync signals when project prop changes
   createEffect(() => {
     const p = props.project;
-    if (!p) return;
+    if (!p) {
+      setConfirmRemoveProjectId(null);
+      return;
+    }
+    setConfirmRemoveProjectId(null);
     setName(p.name);
     setSelectedHue(hueFromColor(p.color));
     setBranchPrefix(sanitizeBranchPrefix(p.branchPrefix ?? 'task'));
@@ -69,6 +81,10 @@ export function EditProjectDialog(props: EditProjectDialogProps) {
   }
 
   const canSave = () => name().trim().length > 0;
+  const confirmRemoveTaskCount = () => {
+    const projectId = confirmRemoveProjectId();
+    return projectId ? getProjectTaskCount(store, projectId) : 0;
+  };
 
   function handleSave() {
     if (!canSave() || !props.project) return;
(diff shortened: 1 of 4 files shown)


johannesjo:
## Review — Approve ✅

Clean, well-scoped fix for #200. The missing-folder banner's **Remove** now routes through the same `ConfirmDialog` the Sidebar uses, and the task-count + copy logic is extracted into a testable `project-remove-confirmation.ts` shared by both call sites. No correctness bugs found.

### Verified locally (all passing)
- `tsc --noEmit`, ESLint, and Prettier — clean on all changed files
- Confirmation copy is **byte-identical** to the prior Sidebar strings → no wording regression
- DRY refactor is **complete** — no leftover inline task-count/copy duplicates at either site
- Behavioral equivalence of the two paths holds: EditProjectDialog always calls `removeProjectWithTasks`, which with 0 tasks falls through to `removeProject` (whose guard also no-ops if tasks still linked) — same result as the Sidebar's `taskCount === 0 → removeProject` branch
- Stacked `ConfirmDialog`-over-`EditProjectDialog` is a supported pattern here: `Dialog` uses a `Portal` + global dialog-stack, so Escape / `aria-modal` / backdrop-click all correctly target the topmost dialog
- Helper logic exercised directly (6/6 assertions incl. unknown-project → 0). Note: `vitest` couldn't run in my sandboxed worktree (read-only symlinked `node_modules`), so I transpiled + asserted the functions out-of-band rather than via `vitest run` — worth a maintainer confirming `npm run test` locally.

### Non-blocking notes
1. **`onClose` fires even when removal is aborted.** `removeProjectWithTasks` resolves normally when it *keeps* the project (a task failed to close → `hasRemainingTasks`), so `.then(onClose)` closes the dialog and signals success anyway. This is **pre-existing** (the old `await …; props.onClose()` had the same shape) and the Sidebar path behaves equivalently, so it's consistent — just not improved here. Out of scope; flagging only.
2. **`createEffect` sets `setConfirmRemoveProjectId(null)` in both branches** — could hoist to one call before `if (!p) return` for clarity.
3. **`confirmRemoveTaskCount()` recomputes twice per render** (once for `message`, once for `confirmLabel`), each iterating all tasks; the Sidebar computes it once via an IIFE. Negligible perf, minor inconsistency.
4. **Tests cover the pure helpers only** — the actual behavioral fix (Remove now opens a confirm instead of deleting immediately) isn't covered. Reasonable given the low risk, but a small interaction test ("click Remove → ConfirmDialog opens → confirm → `removeProjectWithTasks` called") would lock in the regression this fixes.

### Nice
Extracting into a dependency-light pure module (`ProjectTaskCountSource` decouples from the full store) is a genuine testability win and removes real duplication. Small, focused, matches existing conventions.

_Reviewed at 4811f9b._

LarryHu0217:
Thanks for the detailed review. I ran the focused helper test plus the broader local checks before opening this: `npm run test -- src/components/project-remove-confirmation.test.ts`, `npm run typecheck`, `npm run check`, and `npm run test`. The non-blocking notes make sense to me; happy to follow up if you want the interaction test added here or in a separate PR.

LarryHu0217:
I checked the unstable status here: the only GitHub Actions run is CI #28874846464 with conclusion `action_required` and no logs/jobs, and this PR does not touch workflow files. I also tried rerunning it, but GitHub does not allow it from my side. So I think this only needs maintainer-side approval/rerun of the workflow; no code change appears needed.

johannesjo:
Thanks for this, @LarryHu0217 — and for the thorough test + write-up. 🙏

I ended up shipping the #200 fix a bit differently: instead of sharing the copy/count helpers and keeping a `<ConfirmDialog>` at each call site, I extracted the whole thing into a single shared `RemoveProjectConfirm` component, so both the Sidebar and the missing-folder banner reduce to one element with a single `onConfirm` (no duplicated dialog wiring to keep in sync).

I did keep your `getProjectTaskCount` helper and its test — that open+collapsed counting is the one bit worth locking down, and `RemoveProjectConfirm` now uses it.

Closing as superseded, but the fix (and your test) live on. Thanks again!

LarryHu0217:
Thanks for explaining. The shared `RemoveProjectConfirm` direction makes sense, and I am glad the task-count helper/test was useful.

Issue #214, opened 2026-07-07
UX polish batch: nine small verified fixes


johannesjo:
All small, independently fixable, and verified against the code. Checklist form so they can be picked off one at a time (several are good first issues):

- [ ] **Setting label says "on merge" but acts on close.** "Always delete branch and worklog on merge" (`EditProjectDialog.tsx:348`) toggles `deleteBranchOnClose`, whose only readers are close paths (`tasks.ts:417`, `TaskPanel.tsx:598`, `CloseTaskDialog.tsx:99`); the merge path uses a separate `cleanupAfterMerge` checkbox. Also "worklog" is presumably "worktree".
- [ ] **Worktree path clips without ellipsis.** `TaskBranchInfoBar.tsx:247` renders the full path inside a nested `inline-flex` `<button>`; the InfoBar's declared `text-overflow: ellipsis` (`InfoBar.tsx:33-35`) can't apply to a nested flex child, so long paths clip hard. Abbreviate with the `~`-shortening already used in the sidebar and put the full path in the tooltip.
- [ ] **Dead overlay rule.** `.dialog-overlay` background in `styles.css:1363` (44% black) is always overridden by the inline `rgba(0,0,0,0.55)` in `Dialog.tsx:94` — they disagree by 11%; delete one.
- [ ] **Import continues after Cancel.** `ImportWorktreesDialog` Cancel is never disabled (`:305-317`); the sequential `createImportedTask` loop (`:113-121`) keeps running after close, reporting into an unmounted dialog. Disable Cancel while importing or toast on background completion.
- [ ] **Arena swallows Escape.** `App.tsx:685` `if (store.showArena) return;` — wire Escape to close the arena (or its internal back navigation).
- [ ] **Keybinding-migration banner reflows the whole app.** Injected between titlebar and main (`App.tsx:818-886`), so every task column and terminal reflows when it appears/dismisses. Make it a toast/dialog instead.
- [ ] **Silent no-op shortcuts.** `mergeTask`/`pushTask` early-return without feedback on precondition failure (`tasks.ts:566-571, 613-616`); the keyboard handlers (`App.tsx:663-670`) don't pre-check git isolation, so merge/push shortcuts on a non-worktree task do nothing silently. Show a brief notification.
[...]

PR #218, opened 2026-07-07
Polish small UX issues from audit

LarryHu0217:
## Summary
- Fix the close-cleanup setting copy so it says the branch/worktree cleanup happens on close, not on merge.
- Move the existing home-path abbreviation into a shared helper, reuse it for the task worktree path, and keep the full path in the tooltip with ellipsis support.
- Remove the stale `.dialog-overlay` background rule that was overridden by the inline dialog overlay background.
- Let the Escape close-dialog action close Arena through the same reset-and-close path as the Arena close button.

Addresses part of #214.

## Validation
- `npm run test -- src/lib/path.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run compile`
- `npm run test`

The local git hooks also reran `npm run check` and `npm run test` during commit/push.


johannesjo:
Thank you very much! <3

PR #219, opened 2026-07-07
Confirm before discarding new task drafts

### LINKED ISSUE #201: Overlay-click/Escape silently discards the NewTaskDialog prompt draft
## Problem

A misclick on the dialog overlay (or Escape) silently discards a possibly multi-paragraph prompt typed into NewTaskDialog.

- Dialog state is unconditionally reset on next open: `setPrompt(''); setName(''); …` (`NewTaskDialog.tsx:185-200`).
- The shared Dialog closes on overlay click (`Dialog.tsx:97-99`) and Escape (`:40-50`); there is no dirty-check anywhere.

The dialog panel is 560 px wide — everything around it is a discard button. The same applies to unsaved EditProjectDialog changes and a hand-written squash message in MergeDialog.

## Proposed fix

Minimum: don't close NewTaskDialog on overlay-click/Escape when `prompt().trim()` is non-empty (or show a confirm). Better: persist the draft across accidental closes so reopening restores it — the app already persists per-task prompt drafts, so the pattern exists.

---

LarryHu0217:
## Summary
- Add a NewTaskDialog close guard when the prompt or task name has draft text.
- Show a discard confirmation for overlay click, Escape, and Cancel instead of silently closing.
- Keep the draft in place when the user chooses to keep editing.

Fixes #201.

## Validation
- `npm run test -- src/components/new-task-draft.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run compile`
- `npm run test`

The local git hooks also reran `npm run check` and `npm run test` during commit/push.

### DIFF
--- a/src/components/NewTaskDialog.tsx
+++ b/src/components/NewTaskDialog.tsx
@@ -9,6 +9,7 @@ import {
   untrack,
 } from 'solid-js';
 import { Dialog } from './Dialog';
+import { ConfirmDialog } from './ConfirmDialog';
 import { errMessage } from '../lib/log';
 import { invoke } from '../lib/ipc';
 import { IPC } from '../../electron/ipc/channels';
@@ -54,6 +55,7 @@ import {
   MAX_COORDINATOR_CONCURRENT_TASKS,
   MIN_COORDINATOR_CONCURRENT_TASKS,
 } from '../lib/coordinator-limits';
+import { hasNewTaskDraft } from './new-task-draft';
 
 interface NewTaskDialogProps {
   open: boolean;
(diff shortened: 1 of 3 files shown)


johannesjo:
Thanks for tackling this, @LarryHu0217 — and sorry for the collision. 🙏

It turns out an equivalent fix for #201 landed on `main` independently in 70c9cd7 (`fix(new-task): confirm before discarding a typed prompt on close`) shortly before this PR. It takes the same approach — a `ConfirmDialog` gated behind a `requestClose()` guard on `NewTaskDialog.tsx` — so unfortunately this PR now conflicts in that file and would end up duplicating the feature.

Two notes on why I'm keeping the already-merged version as the base:

- It snapshots the prompt right after open/prefill and only guards when you've typed *beyond* it. `hasNewTaskDraft()` triggers on any non-empty field, which would pop the discard confirmation on dialogs that were **prefilled but untouched** (drag-drop of a GitHub URL → `review <url>`, the arena "Compare results" flow, etc.).

- Your version did catch one real gap the merged one missed, though: it also guards the **task name** field, whereas the merged code only watched the prompt. Good catch — I've folded that into the existing implementation (snapshotting `initialName` alongside `initialPrompt` so it stays prefill-safe).

Going to close this as already-addressed by 70c9cd7 plus that follow-up, but I really appreciate you digging into #201. 🙌

johannesjo:
Closing as already-addressed: #201 was fixed on `main` in 70c9cd7, and the name-field guard this PR added has been folded into that implementation (prefill-safe). Thanks again, @LarryHu0217 🙏

LarryHu0217:
Thanks for the detailed explanation. Closing makes sense since the main-branch fix landed independently, and I appreciate you folding in the name-field guard.

Issue #208, opened 2026-07-07
Add a shared icon module (75 inline SVGs across 31 files, close icon pasted 10x)


johannesjo:
## Problem

There is no icon module: **75 inline `<svg>` blocks across 31 files**, with heavy duplication and mixed media.

- The Octicon close path (`M3.72 3.72…`) is hand-pasted **10 times**; copy/branch and external-link paths several times each; the folder icon exists in two subtly different variants that render visibly differently.
- Text glyphs serve as icons alongside SVGs: `✕` (`SubTaskStrip.tsx:74`, `CommitTreeOverlay.tsx:149`, `TerminalSearchOverlay.tsx:150`), `✓` (`SubTaskStrip.tsx:236`, `TaskStepsSection.tsx:231`), `❯` throughout `PromptInput.tsx`, and an emoji `📁` (`NewTaskDialog.tsx:1093`) — emoji render platform-dependently and full-color, breaking the otherwise monochrome `currentColor` icon language.
- Icon sizes span ~9 distinct widths (8–16 px).

## Proposed fix

A single `src/components/icons.tsx` exporting the ~20 actual glyphs as components with a `size` prop (16 default, 12 small), then replace the pasted SVGs and glyph/emoji icons. Mechanical work, good first issue.

---
_From a full design/UX audit (2026-07-07); adversarially verified (counts re-derived). Line numbers as of b342bbd (v1.12.0)._

truongthinh1412:
Hi @johannesjo! 👋

I hope you don't mind me jumping in — I came across #208 while looking for a way to contribute to parallel-code, and it seemed like a great fit. I've put together a PR that completes the shared icon module migration: **#276** extends `src/components/icons.tsx` to cover the remaining glyphs and replaces the leftover inline `<svg>` blocks and ✓/× text-glyph icons, keeping the rendering pixel-identical everywhere.

If the scope or approach isn't quite what you had in mind — or a 39-file PR feels like too much in one go — I'd be more than happy to split it into smaller tranches, tweak the component API, or rework anything to better match the project's direction. Just say the word. 🙂

Thanks for maintaining such a nice project, and for labeling approachable issues like this one — it genuinely makes it much easier for newcomers to find a way in!

PR #220, opened 2026-07-07
Refactor duplicated inline icons into shared components

LarryHu0217:
## Summary
- add a shared `src/components/icons.tsx` module with reusable `currentColor` icon components
- replace the duplicated Octicon close path call sites with `CloseIcon`
- replace the named glyph/emoji examples from #208 (`✕`, `✓`, `📁`) with shared SVG icons
- reuse shared folder, branch, and commit-graph icons in the touched dialog/overlay surfaces

Refs #208

## Validation
- `npm run typecheck`
- `npm run check`
- pre-push hook: `npm run check` + `npm run test`


johannesjo:
Thank you very much! <3

Issue #158, opened 2026-05-30
[Quality audit] Deduplicate backend IPC, Git, remote, and request lifecycle code


johannesjo:
Derived from the local audit report `CODE_QUALITY_FINDINGS.md` created on 2026-05-30.

## Scope

This issue groups backend duplication and complexity opportunities outside the MCP coordinator itself.

## Findings

- [ ] Finding 4: duplicated IPC channel source of truth.
  - Evidence: `electron/preload.cjs`, `electron/ipc/channels.ts`, `electron/main.ts`, `electron/preload-allowlist.test.ts`.
  - Fix direction: generate the preload CJS channel list from `IPC`, or move both to a generated shared manifest with exact-set tests.
- [ ] Finding 5: remote HTTP routing is a monolithic auth-sensitive handler.
  - Evidence: repeated coordinator task guards in `electron/remote/server.ts`.
  - Fix direction: extract `jsonReply`, `readBody`, `requireOwnedTask`, and split coordinator API routes into smaller handlers or a route table.
- [ ] Finding 8: repeated IPC validation boilerplate in Git handlers.
  - Evidence: repeated `baseBranch`, `projectRoot`, `worktreePath`, and `branchName` validation in `electron/ipc/register.ts`.
  - Fix direction: add typed helpers such as `optionalBaseBranch(args)` and wrappers for common Git handler shapes.
- [ ] Finding 11: `.git/info/exclude` append logic is duplicated.
  - Evidence: `electron/ipc/register.ts`, `electron/ipc/steps.ts`, and `electron/ipc/git.ts`.
  - Fix direction: extract `resolveGitInfoExcludePath(worktreePath)` and `appendGitInfoExcludeBlock(...)`.
- [ ] Finding 12: changed-file assembly repeats across Git IPC paths.
  - Evidence: several `ChangedFile[]` rebuild paths in `electron/ipc/git.ts`.
  - Fix direction: extract `changedFilesFromMaps(...)` and `countReadableTextLines(...)`.
- [ ] Finding 13: `spawnAgent` mixes too many responsibilities.
  - Evidence: `electron/ipc/pty.ts:176-501`.
  - Fix direction: extract env/spec construction, PTY output handler attachment, and session cleanup.
- [ ] Finding 14: ask-code providers duplicate request lifecycle.
[...]

PR #221, opened 2026-07-08
refactor: deduplicate backend IPC and Git flows

ASRagab:
## Summary

Addresses #158 by deduplicating backend IPC/Git helper paths and related request lifecycle code.

## Review

Adversarial review ran before final validation. Fixable follow-up was repaired before the final gate pass.

## Verification

- [x] `npm run check`
- [x] `npm run lint:arch`
- [x] `npm run lint:dead`
- [x] `npm test` (`83 passed | 2 skipped`, `1508 passed | 24 skipped`)
- [x] `npm run dev` smoke reached Vite/Electron readiness at `http://localhost:1421/` and cleaned up
- [x] pre-push hook reran `npm run check` and `npm test`

Refs #158


johannesjo:
## Review

Backend refactor reviewed as *no behavior change*. I checked the cross-cutting channel/type changes directly and ran deep dives on the ask-code lifecycle, git helpers, and pty/remote-server, plus empirical git testing. **One blocker, otherwise a clean and faithful refactor** (it even folds in a few latent bugfixes).

---

### 🔴 Blocker — sandbox & symlink git-excludes write to a file git never reads (broken for every worktree)

**Where:** `electron/ipc/git-exclude.ts` `resolveLocalGitInfoExcludePath`, consumed by `ensureSandboxExcludes` and `ensureSymlinkExcludes` in `electron/ipc/git.ts`.

**What changed:** On `main`, both flows resolved the exclude path via `git rev-parse --git-common-dir` → the **common** `<repo>/.git/info/exclude`. The new shared `resolveGitInfoExcludePath` is **local-first**: for a linked worktree (`.git` is a *file* → `gitdir: …/.git/worktrees/<name>`) it returns the **per-worktree** `…/.git/worktrees/<name>/info/exclude` and never reaches the `--git-common-dir` fallback.

**Why it's wrong:** Git reads `info/exclude` only from the *common* dir. Verified empirically on git 2.43.0 in a real linked worktree:

```
write .git/worktrees/<wt>/info/exclude → git status: "?? probe"   (NOT excluded)
write .git/info/exclude                → git status: (empty)       (excluded)
git check-ignore -v probe → .git/info/exclude:20:probe   ← attributed only to the common dir
```

**Impact:** This app *always* creates linked worktrees (`<root>/.worktrees/<branch>`), so `.git` is always a file. On every agent spawn the bwrap char-device placeholders (`ensureSandboxExcludes`) and the `/node_modules`, `/.env` symlink patterns (`ensureSymlinkExcludes`) now land in a dead file → they **reappear in `git status` / `git ls-files` / the diff view** and can be swept into a "commit all". Both functions become silent no-ops for their entire purpose (`appendGitInfoExcludeBlockAtPath` even `mkdirSync`s the wrong `info/` dir, so nothing errors).

**Why CI is green:** `git-symlink-excludes.test.ts` never mocks `fs.statSync`, so `statSync('/worktree/.git')` throws `ENOENT` and every test falls through to the mocked `rev-parse` — the new fast-path is never exercised. And `git-helpers.test.ts` (`resolves the exclude path for linked worktrees`) *asserts the per-worktree path as correct*, codifying the bug.

**Suggested fix:** Make `resolveGitInfoExcludePath` resolve the **common** dir. Simplest is to drop `resolveLocalGitInfoExcludePath` and always `git rev-parse --git-common-dir` (matches `main` exactly; that branch already exists in the fallback). If you want to keep an FS fast-path, for the `.git`-file case read `<gitdir>/commondir` and resolve `<gitdir>/<commondir>/info/exclude`, not `<gitdir>/info/exclude`. Bonus: this also fixes a *pre-existing* latent bug in `steps.ts`, whose old code used the per-worktree path too (so `.claude/steps.json` exclusion was already ineffective in worktrees). Recommend adding a test that runs real `git status`/`check-ignore` against a linked worktree rather than asserting the resolved path string.

---

### 🟢 Everything else checks out

- **request-registry / ask-code** — behavior-preserving and actually *fixes* a latent bug: on `main`, replacing a same-`requestId` request let the old process's `close` handler delete the *new* request's map entry and clear its timeout; the new `token` guard makes `finish()` correctly no-op for a stale handle. "Exactly one `done`" holds on every path, no slot leaks, no stale-timer race. Limits/timeout/message text unchanged. Good coverage in the new `request-registry.test.ts`.
- **pty.ts / remote server.ts** — no behavioral delta, no security regression: all 13 remote routes map 1:1, token-class gating / ownership checks / `timingSafeEqual` intact, `ENV_BLOCK_LIST` and docker spawn args byte-identical. Two latent bugfixes: multi-byte UTF-8 prompts split across TCP chunks no longer corrupt (`Buffer.concat`), and the oversize-body `413`/`400` now reaches the client instead of a socket reset.
- **register.ts** — channel wiring identical: 107 `IPC.*` handler refs, empty set-diff vs `main`; no channel dropped, renamed, or rewired.
- **channels enum → JSON manifest** — channel *values* byte-identical (139/139). Build-safe (Electron 40 → Node 22 supports ESM JSON import attributes; `tsc` copies the JSON to `dist-electron`; Vite handles it). Preload allowlist set **unchanged** (no security-boundary widening), and sync is now enforced by a test plus an improved runtime check that also catches *extra* channels.
- **install.sh** — latent bugfix folded in: the macOS package build now also runs `build:remote` + `build:mcp`.

### 🟡 Minor / nits (non-blocking)

- Enum key renames (`TaskCreated` → `MCP_TaskCreated`, `CreateTaskRequest` → `Remote_CreateTaskRequest`, ~22 keys) are broader than "dedup" implies, and the `Foo_Bar` style is inconsistent with the `PascalCase` of the other keys.
- Section comments were lost when `channels`/`preload.cjs` were flattened to the manifest (partly mitigated by the new key prefixes).
- Brittle string-matching tests: `install-script.test.ts` asserts the whole command line verbatim; preload/manifest sync relies on regex-scraping `preload.cjs`.
- `git-exclude.ts` marker dedup changed from substring `includes` to an exact trimmed-line match — harmless given markers are always on their own line.
- `ensureSandboxExcludes` now labels non-ENOENT read failures as `"Failed to append…"` instead of `"Failed to read…"` — still surfaced, just mislabeled.

---

*Confidence in the blocker: ~97% — verified by reading old-vs-new for every hunk, confirming both flows route through the local-first helper, and empirically testing git's exclude behavior (`git status` + `check-ignore -v`) on a real linked worktree at git 2.43.0.*

ASRagab:
Addressed the linked-worktree exclude blocker. resolveGitInfoExcludePath now always uses git rev-parse --git-common-dir, so sandbox, symlink, and steps exclusions write to the common info/exclude file Git reads. Added a real linked-worktree regression test covering check-ignore and status. Local full suite passes (83 files, 1511 tests), pre-push checks pass, and GitHub CI is green. Head: e441df7.

johannesjo:
Thank you very much! <3

PR #224, opened 2026-07-09
fix(changed-files): focus open editor button clicks

FourWindff:
## Summary

  Fixes focus behavior when clicking the external open button in the Changed Files panel.

  This addresses two cases:

  1. When the Notes area is focused, clicking a file's external open button in Changed Files should move the highlight to Changed Files only.
  2. When the TaskPanel is not focused, clicking a file's external open button in its Changed Files panel should select that TaskPanel and highlight Changed Files.

  ## Changes

  - Adds an `onOpenInEditorClick` callback to `ChangedFilesList`.
  - Calls the Changed Files focus path before opening a file in the editor.
  - Updates `TaskChangedFilesSection` so Changed Files focus also activates the task via `setActiveTask`.
  - Keeps `stopPropagation()` on the external open button so clicking it does not also trigger the file row diff click.

### DIFF
--- a/src/components/ChangedFilesList.tsx
+++ b/src/components/ChangedFilesList.tsx
@@ -20,6 +20,7 @@ interface ChangedFilesListProps {
   onFileClick?: (file: ChangedFile) => void;
   /** Optional path to visually mark as the active/open diff target. */
   activeFilePath?: string | null;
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #226, opened 2026-07-14
fix(commits): stop polling hidden task panels

### LINKED ISSUE #205: Every mounted task panel polls GetBranchCommits every 5s regardless of visibility
## Problem

Every mounted task panel polls `GetBranchCommits` on a fixed 5-second interval regardless of visibility or activity — and every task stays mounted, even hidden ones.

- `TaskPanel.tsx:198-234`: `setInterval(fetchCommits, 5000)` per panel; only guards are `isLandedTask()`/git-isolation, not visibility.
- All tasks stay mounted via the panel cache (`TilingLayout.tsx:233-244`); focus mode hides inactive panels with `visibility: hidden` rather than unmounting (`:581`).
- At 30 tasks that's ~6 git subprocess invocations per second just for commit lists, plus 30 live xterm instances.

The codebase already solved this problem for git *status* polling — `computeAllTasksInterval` scales the interval with task count (`src/store/taskStatus.ts:1071-1074`) — the commit poll just never got the same treatment. Viewport visibility is also already computed and stored (`taskViewportVisibility`, `TilingLayout.tsx:98-113`).

## Proposed fix

- [ ] Gate the per-panel commit poll on `taskViewportVisibility` (skip or slow polling for offscreen/hidden panels), and/or reuse the scaled-interval pattern from `taskStatus.ts`.
- [ ] Longer term: consider unmounting panel bodies far offscreen — terminals already survive remounts via DOM reparenting (`TaskPanel.tsx:247-265`) and the pty buffer lives in the main process.
- [ ] Sidebar task list is a plain `<For>` (`Sidebar.tsx:717`) — fine for now, but worth virtualizing if task counts grow.

---

LarryHu0217:
## Summary

- stop the five-second branch-commit poll while a tiled task panel is offscreen
- only poll the active task panel in focus mode
- refresh immediately when a task panel becomes visible again
- cover tiled and focus-mode visibility decisions with unit tests

Closes #205

## Test plan

- `npm run check`
- `npm run check:static`
- `npm test` (92 passed, 2 skipped; 1564 tests passed, 23 skipped)

### DIFF
--- a/src/components/TaskPanel.tsx
+++ b/src/components/TaskPanel.tsx
@@ -41,6 +41,7 @@ import { isMac } from '../lib/platform';
 import type { Task } from '../store/types';
 import type { CommitInfo } from '../ipc/types';
 import { isLandedTaskState } from '../store/landing';
+import { shouldPollTaskCommits } from './task-commit-polling';
 
 interface TaskPanelProps {
   task: Task;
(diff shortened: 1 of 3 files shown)


johannesjo:

PR #227, opened 2026-07-17
fix(theme): adapt light theme contrast

### LINKED ISSUE #207: Light theme: dark-only alpha overlays, hardcoded diff/search colors, contrast-tooling gaps
## Problem

Dark-theme assumptions are hardcoded in TSX and break on the light preset (islands-light); a few gaps in the contrast tooling let this go unnoticed.

**White/black alpha literals that don't adapt:**
- "Subtle lift" hovers built from white alpha are invisible on white: `DiffViewerDialog.tsx:367` (`rgba(255,255,255,0.06)`), `ReviewSidebar.tsx:51`, `AskCodeCard.tsx:106`, `ScrollingDiffView.tsx:605`, `Dialog.tsx:121` (inset highlight).
- `TerminalView.tsx:1212` `rgba(0,0,0,0.85)` overlay is a black slab on a light theme.
- Diff add/remove and search-highlight colors are hardcoded: `ScrollingDiffView.tsx:41-42,110-111`; `TerminalView.tsx:175-178` (`#ffd54f`, `#ff8a00`, …).

The fix pattern already exists in the codebase: `color-mix(in srgb, var(--fg) N%, transparent)` appears **89×** in `styles.css` — this is just finishing the migration for the values that live in TSX.

**Tooling gaps:**
- `checkThemeContrast` (`src/lib/custom-theme.ts:264-269`) checks `--fg-muted` but not `--fg-subtle`, which is used for real content (timestamps, placeholders, tertiary labels); classic's `--fg-subtle` is **2.73:1** on its elevated bg.
- The existing preset contrast test (`src/lib/custom-theme-contrast.test.ts`) is a soft audit (`expect(warnings).toBeDefined()` — never fails). Consider making regressions gate.
- The terminal background per preset is a hand-maintained 12-entry hex table duplicating `--task-panel-bg` (`src/lib/theme.ts:46-59`) — any preset edit must be made twice; derive it from the CSS var instead (a `readCssVarsForPreset` helper already exists).

## Proposed fix

- [ ] Replace white/black alpha literals in TSX with `color-mix(... var(--fg)/var(--bg) ...)` equivalents.
- [ ] Tokenize diff add/remove and search-highlight colors (`--diff-add-bg`, `--diff-remove-bg`, `--search-match`) with light-preset values.
- [ ] Add `--fg-subtle` to `CONTRAST_PAIRS` and lift the failing preset values.
- [ ] Make the preset contrast test assert (fail on regressions) rather than just warn.
- [ ] Derive terminal bg from `--task-panel-bg` instead of the parallel hex table.

---
_From a full design/UX audit (2026-07-07); adversarially verified — an earlier stronger claim ("presets fail the checker's own rule") was refuted and narrowed to the gaps above. Line numbers as of b342bbd (v1.12.0)._

ASRagab:
## Summary

- Replace dark-only overlay and inset colors with theme-adaptive surfaces.
- Add semantic diff and search tokens, including light-preset values and xterm resolution.
- Enforce `--fg-subtle` contrast and make the built-in preset audit a hard gate with inherited root tokens.
- Derive built-in terminal and cloned-theme backgrounds from `--task-panel-bg` instead of a duplicate table.

## Root cause

Several renderer styles and terminal decorations bypassed the theme token system, while the contrast audit omitted tertiary text and could never fail. Built-in terminal backgrounds also duplicated CSS values in TypeScript, allowing the two sources to drift.

## Validation

- `npm run check`
- `npm test -- --reporter=dot` — 1,571 passed, 24 skipped
- Electron development smoke (`npm run dev`)
- Islands Light click-through: Settings/Themes rendering, themed terminal search highlight, New Task draft retention after Escape/Cancel, and task-count-aware project removal confirmation

Fixes #207.

### DIFF
--- a/src/components/AskCodeCard.tsx
+++ b/src/components/AskCodeCard.tsx
@@ -103,7 +103,7 @@ export function AskCodeCard(props: AskCodeCardProps) {
           'justify-content': 'space-between',
           padding: '4px 10px',
           'border-bottom': `1px solid ${theme.borderSubtle}`,
-          background: 'rgba(255,255,255,0.03)',
+          background: 'color-mix(in srgb, var(--fg) 3%, transparent)',
         }}
       >
         <span
--- a/src/components/Dialog.tsx
+++ b/src/components/Dialog.tsx
@@ -118,7 +118,8 @@ export function Dialog(props: DialogProps) {
               'flex-direction': 'column',
               gap: '16px',
               outline: 'none',
-              'box-shadow': '0 12px 48px rgba(0,0,0,0.5), 0 0 0 1px rgba(255,255,255,0.03) inset',
+              'box-shadow':
+                '0 12px 48px rgba(0,0,0,0.5), 0 0 0 1px color-mix(in srgb, var(--fg) 3%, transparent) inset',
               ...props.panelStyle,
             }}
             onClick={(e) => e.stopPropagation()}
--- a/src/components/DiffViewerDialog.tsx
+++ b/src/components/DiffViewerDialog.tsx
@@ -365,7 +365,7 @@ function DiffViewerContent(props: DiffViewerDialogProps) {
           value={searchQuery()}
           onInput={(e) => setSearchQuery(e.currentTarget.value)}
           style={{
-            background: 'rgba(255,255,255,0.06)',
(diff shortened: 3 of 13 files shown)


johannesjo:
Thank you very much! <3

PR #228, opened 2026-07-18
fix(import): prevent dismissal while importing worktrees

LarryHu0217:
## Summary

- Route Escape, overlay clicks, and the Cancel button through a shared close guard.
- Disable and visually mute Cancel while sequential worktree imports are running.
- Preserve the existing successful-completion close path.

Addresses one remaining item from #214.

## Validation

- `npm test -- src/components/import-worktrees-dismiss.test.ts` (2 passed)
- `npm run check`
- `npm run lint:dead`
- `npm run lint:arch`
- `npm test` (1573 passed, 23 skipped)

### DIFF
--- a/src/components/ImportWorktreesDialog.tsx
+++ b/src/components/ImportWorktreesDialog.tsx
@@ -5,6 +5,7 @@ import { invoke } from '../lib/ipc';
 import { IPC } from '../../electron/ipc/channels';
 import { createImportedTask, getProjectPath, loadAgents, store } from '../store/store';
 import { theme } from '../lib/theme';
+import { createImportWorktreesCloseHandler } from './import-worktrees-dismiss';
 import type { Project } from '../store/types';
 import type { AgentDef, ImportableWorktree } from '../ipc/types';
 
@@ -22,6 +23,7 @@ export function ImportWorktreesDialog(props: ImportWorktreesDialogProps) {
   const [loading, setLoading] = createSignal(false);
   const [importing, setImporting] = createSignal(false);
   const [error, setError] = createSignal('');
+  const requestClose = createImportWorktreesCloseHandler(importing, () => props.onClose());
 
   const trackedWorktreePaths = createMemo(() => {
     const projectId = props.project?.id;
@@ -127,7 +129,7 @@ export function ImportWorktreesDialog(props: ImportWorktreesDialogProps) {
   }
 
   return (
-    <Dialog open={props.open} onClose={props.onClose} width="560px" panelStyle={{ gap: '18px' }}>
+    <Dialog open={props.open} onClose={requestClose} width="560px" panelStyle={{ gap: '18px' }}>
       <div style={{ display: 'flex', 'flex-direction': 'column', gap: '18px' }}>
         <div>
           <h2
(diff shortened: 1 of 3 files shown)


johannesjo:
Thanks for picking this one up. The wiring is right, and I appreciate that you spotted the success-path close at line 121 and deliberately left it bypassing the guard — that's the easy trap here and you avoided it.

Before merging though, I want to check the premise, because I don't think this guard can ever fire.

`createImportedTask` (`src/store/tasks.ts:419-455`) is `async` in signature only — there's no `await` anywhere in its body. `initTaskInStore` is a synchronous `setStore(produce(...))`, `markAgentSpawned` and `rescheduleTaskStatusPolling` are both synchronous, and `saveState()` on line 453 is called floating (unawaited). So the loop in `ImportWorktreesDialog.tsx:114-120` resolves entirely in microtasks, and microtasks drain before the browser paints and before it dispatches queued input. Which means:

- `disabled={importing()}` and `opacity: 0.4` on Cancel are never painted — there's no frame in which anyone can see or click the muted button.
- Escape and overlay clicks can't be delivered while `importing()` is true either. If a large import blocks the main thread, the queued input fires *after* the loop finishes, when `importing()` is already back to `false`.

That's my mistake in the issue text, not yours — I wrote "keeps running after close, reporting into an unmounted dialog" in #214 from a static read, and with a synchronous loop that window doesn't exist. Even the residual case is benign: `props.onClose()` runs, then `finally { setImporting(false) }` writes a signal on a disposed Solid owner, which is a no-op in Solid (no warning, no crash).

So two ways forward, and I'd rather pick one explicitly than merge on the assumption a live bug was fixed:

1. **Keep it as future-proofing.** Reasonable if `createImportedTask` is likely to gain a real `await` later (awaiting `saveState`, an IPC worktree registration). But then it wants a comment saying that's what it is — and an escape hatch. A hard block on all three dismissal paths means the day that loop *does* become genuinely async, a hung call traps the user in a modal with no Escape, no overlay click, no Cancel, and no progress beyond a static "Importing…". A progress count (`Importing 3 of 12`) or the other option from the issue (allow dismissal, continue in background, toast on completion) degrades much better.
2. **Drop this one and take a different item from #214.** Several of the remaining ones are more clearly observable.

Two smaller notes either way:

- `import-worktrees-dismiss.ts` is a whole module plus test file for `if (!isImporting()) onClose()`, with the caller's name baked into the export. We do have a sibling-helper convention (`project-remove-confirmation.ts`, `prompt-draft.ts`), but those hold actual logic. An inline `const requestClose = () => { if (!importing()) props.onClose(); }` in the component is shorter and keeps the guard next to the signal it reads.
- Relatedly, the two tests assert that an `if` statement works — they can't fail for any reason related to this feature. The behaviour worth covering is the wiring: that `Dialog`'s `onClose` receives `requestClose`, and that line 121 deliberately does *not*. That's where a regression would actually land, and it's the bit most likely to get "tidied up" by a future refactor — worth a short comment on line 121 as well.

Not blocking from my side, I just want us to agree on which of the two paths above we're on. I'll fix the wording of that checkbox in #214 regardless, since the behaviour I described there wasn't reachable.

LarryHu0217:
Thanks for tracing the execution path. I agree the reported dismissal window is not reachable today, and keeping a hard guard as speculative future-proofing would be the wrong UX without progress or cancellation.

I am closing this PR rather than merging code for a non-observable bug. I will take a separately observable item from #214 instead.

johannesjo:
Thank you  @LarryHu0217 !

Issue #211, opened 2026-07-07
Reduced-motion guards, opt-in terminal screen-reader mode, tooltip upgrade


johannesjo:
## Problem

**Infinite animations lack `prefers-reduced-motion` guards in the main stylesheet.** `@keyframes statusPulse` drives `.status-dot-pulse` (1.9 s infinite — potentially on many task dots simultaneously) at `styles.css:1413-1424`; `askcode-pulse` (`:2263`); inline `keybind-pulse` (`HelpDialog.tsx:529-535`). `styles.css` has exactly **one** reduced-motion block, covering only `.projects-collapser` (`:977-981`). The arena CSS files all guard properly — the pattern exists; the main file doesn't use it.

**Terminals are silent to screen readers.** No `screenReaderMode` anywhere in src/ — xterm.js ships an off-by-default SR mode, and the app's primary content (agent output) is unreadable by assistive tech. It costs rendering performance, so opt-in via Settings is the right shape.

**All 124 tooltips are native `title=`** with no custom tooltip component: ~1 s hover delay, never shown on keyboard focus, unstyled. `IconButton` funnels through a `title` prop consistently (which at least provides accessible names), so a single upgrade point exists.

## Proposed fix

- [ ] One `@media (prefers-reduced-motion: reduce)` block disabling the pulse/appear/remove keyframe animations in `styles.css`.
- [ ] Settings toggle wiring xterm `screenReaderMode: true` into terminal creation.
- [ ] (Stretch) small custom tooltip shown on hover *and* `:focus-visible`; `aria-label` fallback prop on `IconButton`.

---
_From a full design/UX audit (2026-07-07); adversarially verified. Line numbers as of b342bbd (v1.12.0)._

PR #230, opened 2026-07-18
fix(accessibility): honor reduced motion

LarryHu0217:
## Summary
- disable task enter/exit animations when the OS requests reduced motion
- suppress status, Ask Code, and keybinding pulse animations in reduced-motion mode
- preserve the status dot's busy sub-state with a static outer ring after its pulse is removed
- move the main reduced-motion block to the end of the stylesheet and reserve `!important` for inline animations
- avoid stale task-entry classes when reduced motion is already enabled
- strengthen regression coverage so selector groups are paired with their declarations

Functional loading spinners remain animated because rotation is the loading signal in icon-only states; status dots now have a non-motion fallback.

## Validation
- `npx vitest run src/lib/reduced-motion-styles.test.ts`
- `npm test` (1,573 passed, 23 skipped)
- `npm run check`
- `npm run check:static`

Addresses the reduced-motion item in #211.

### DIFF
--- a/src/components/AskCodeCard.tsx
+++ b/src/components/AskCodeCard.tsx
@@ -149,6 +149,7 @@ export function AskCodeCard(props: AskCodeCardProps) {
       >
         <Show when={loading() && !response()}>
           <span
+            class="askcode-loading-pulse"
             style={{
               color: theme.fgSubtle,
               animation: 'askcode-pulse 1.5s ease-in-out infinite',
@@ -160,6 +161,7 @@ export function AskCodeCard(props: AskCodeCardProps) {
         <Show when={response()}>{response()}</Show>
         <Show when={loading() && response()}>
           <span
+            class="askcode-loading-pulse"
             style={{
               color: theme.accent,
               'font-size': sf(11),
--- a/src/components/HelpDialog.tsx
+++ b/src/components/HelpDialog.tsx
@@ -364,6 +364,7 @@ export function HelpDialog(props: HelpDialogProps) {
                           </button>
                         </Show>
                         <kbd
+                          class="keybinding-recording-pulse"
                           role="button"
                           tabIndex={0}
                           aria-label={
--- a/src/components/HelpDialog.tsx
+++ b/src/components/HelpDialog.tsx
@@ -364,7 +364,7 @@ export function HelpDialog(props: HelpDialogProps) {
                           </button>
                         </Show>
                         <kbd
-                          class="keybinding-recording-pulse"
+                          class="keybinding-key"
                           role="button"
                           tabIndex={0}
                           aria-label={
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -48,6 +48,7 @@ import { abbreviateHomePath } from '../lib/path';
 import { invoke } from '../lib/ipc';
 import { IPC } from '../../electron/ipc/channels';
 import type { ImportableWorktree } from '../ipc/types';
+import { shouldAnimateTaskAppearance } from '../lib/reducedMotion';
 
 const DRAG_THRESHOLD = 5;
 const SIDEBAR_DEFAULT_WIDTH = 240;
@@ -171,9 +172,17 @@ export function TaskRowShell(props: {
   style?: JSX.CSSProperties;
   children: JSX.Element;
 }) {
+  const className = () =>
+    shouldAnimateTaskAppearance()
+      ? props.class
+      : props.class
+          .split(/\s+/)
+          .filter((className) => className !== 'task-item-appearing')
+          .join(' ');
+
   return (
     <div
-      class={props.class}
+      class={className()}
       role={props.role}
       tabIndex={props.tabIndex}
       data-task-index={props.taskIndex}
--- a/src/components/StatusDot.tsx
+++ b/src/components/StatusDot.tsx
@@ -49,7 +49,7 @@ export function StatusDot(props: {
   const isPulsing = () => props.attention === 'active' || props.status === 'busy';
   return (
     <span
-      class={isPulsing() ? 'status-dot-pulse' : undefined}
+      class={isPulsing() ? 'status-dot-pulse status-dot-ring' : undefined}
       title={getDotTooltip(props.status, props.attention)}
       style={{
         display: 'inline-block',
--- a/src/components/TilingLayout.tsx
+++ b/src/components/TilingLayout.tsx
@@ -29,13 +29,9 @@ import { markDirty } from '../lib/terminalFitManager';
 import { theme } from '../lib/theme';
 import { mod } from '../lib/platform';
 import { createCtrlShiftWheelResizeHandler } from '../lib/wheelZoom';
+import { shouldAnimateTaskAppearance } from '../lib/reducedMotion';
 
 const VIEWPORT_EPSILON_PX = 4;
-
-function shouldAnimateTaskAppearance(): boolean {
-  if (typeof window === 'undefined' || typeof window.matchMedia !== 'function') return true;
-  return !window.matchMedia('(prefers-reduced-motion: reduce)').matches;
-}
 
 /** Tiling-layout top-level child. Distinct from `PanelChild` because this
  *  layout owns its own horizontal drag model — fixed placeholders, per-panel
--- a/src/components/TilingLayout.tsx
+++ b/src/components/TilingLayout.tsx
@@ -32,6 +32,11 @@ import { createCtrlShiftWheelResizeHandler } from '../lib/wheelZoom';
 
 const VIEWPORT_EPSILON_PX = 4;
 
+function shouldAnimateTaskAppearance(): boolean {
+  if (typeof window === 'undefined' || typeof window.matchMedia !== 'function') return true;
+  return !window.matchMedia('(prefers-reduced-motion: reduce)').matches;
+}
+
 /** Tiling-layout top-level child. Distinct from `PanelChild` because this
  *  layout owns its own horizontal drag model — fixed placeholders, per-panel
  *  min/max widths, pixel-precise persisted sizes — that doesn't map onto the
@@ -244,6 +249,7 @@ export function TilingLayout() {
           content: () => {
             const task = store.tasks[panelId];
             const terminal = store.terminals[panelId];
+            const appearanceClass = shouldAnimateTaskAppearance() ? 'task-appearing' : undefined;
             // eslint-disable-next-line solid/components-return-once
             if (!task && !terminal) return <div />;
             return (
@@ -252,7 +258,7 @@ export function TilingLayout() {
                 class={
                   task?.closingStatus === 'removing' || terminal?.closingStatus === 'removing'
                     ? 'task-removing'
-                    : 'task-appearing'
+                    : appearanceClass
                 }
                 style={{
                   height: '100%',
@@ -264,6 +270,10 @@ export function TilingLayout() {
                   'box-sizing': 'border-box',
                 }}
                 onAnimationEnd={(e) => {
+                  if (e.animationName === 'taskAppear')
+                    e.currentTarget.classList.remove('task-appearing');
+                }}
+                onAnimationCancel={(e) => {
                   if (e.animationName === 'taskAppear')
                     e.currentTarget.classList.remove('task-appearing');
                 }}
--- a/src/lib/reduced-motion-styles.test.ts
+++ b/src/lib/reduced-motion-styles.test.ts
@@ -3,8 +3,6 @@ import { resolve } from 'path';
 import { describe, expect, it } from 'vitest';
 
 const css = readFileSync(resolve(__dirname, '../styles.css'), 'utf8');
-const helpDialog = readFileSync(resolve(__dirname, '../components/HelpDialog.tsx'), 'utf8');
-const tilingLayout = readFileSync(resolve(__dirname, '../components/TilingLayout.tsx'), 'utf8');
 
 function reducedMotionBlock(): { block: string; end: number } {
   const marker = '@media (prefers-reduced-motion: reduce)';
@@ -34,7 +32,7 @@ function expectRule(block: string, selectors: string[], declaration: RegExp): vo
   expect(block).toMatch(rule);
 }
 
-describe('reduced-motion styles', () => {
+describe('main stylesheet reduced-motion styles', () => {
   it('pairs each main-app selector group with its reduced-motion override', () => {
     const { block, end } = reducedMotionBlock();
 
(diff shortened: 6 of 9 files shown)


johannesjo:
Reviewed locally on `codex/reduced-motion-211`. The change is sound and I'd merge it — but there's one accessibility trade-off worth a decision before it lands, plus a convention suggestion that would let most of the `!important` go away.

## The thing that could have broken, and doesn't

`.task-removing` / `.task-item-removing` are **exit** animations, and killing an exit animation is the classic way to strand nodes in the DOM — if teardown hangs off `animationend`, `animation: none` means it never fires. It doesn't here. Removal is timer-driven:

- `src/store/tasks.ts:597-606` — `setTimeout(..., REMOVE_ANIMATION_MS)`
- `src/store/terminals.ts:91` — same pattern

`REMOVE_ANIMATION_MS` is 300 while the animations are `0.24s`, so the timer already outlived the animation and was never coupled to it. A repo-wide search for `animationend|transitionend|getAnimations|.animate(` turns up exactly one handler — `TilingLayout.tsx:266-269` — and it only strips a cosmetic class. Nothing leaks. Under reduced motion the panel simply sits for 300ms and then vanishes without fading.

## One real trade-off: the status dot loses a distinction

I initially assumed the pulse was redundant encoding (color + ring + tooltip all carry the state anyway). That's **not** right, and it's the one thing here I'd want a deliberate call on.

`getDotColor` and `getDotShadow` (`StatusDot.tsx:6-28`) key off `attention` and return early; `getDotTooltip` (`:30-41`) likewise early-returns on `attention`. But `isPulsing()` (`:49`) is `attention === 'active' || status === 'busy'`. So once `attention` is anything other than `active`, the `busy` vs non-`busy` **`status`** changes nothing except the pulse.

Two collisions that are reachable and fully identical once the pulse is gone — same color, same ring, same tooltip:

- **`attention='needs_input'` + `busy` vs `waiting`.** `getTaskAttentionState:844-845` returns `needs_input` on `hasQuestion` regardless of activity, while `getTaskDotStatus:882-885` returns `busy` on `hasActive`. An agent that's asked a question but is still inside the idle window satisfies both.
- **`attention='error'` + `busy` vs `waiting`.** `hasTaskAgentError` and `hasRunningTaskActivity` read different agents, so one errored agent alongside one running agent hits both.

(There's a third, `attention='review'` + `busy` vs `review` — the two functions check `needsReview` and `hasActive` in opposite order. Color and ring are identical there, but the tooltip *does* differ, so it's the mild case.)

The important state survives: `attention='active'` maps to `theme.accent`, which nothing else uses, so "is this agent working" is still color-encoded at the top level. What's lost is the "…and it's also still churning" sub-state inside `needs_input` / `error`. And `title` is hover-only, so it's not a real substitute for a visual encoding.

I don't think this blocks the PR — shedding motion-encoded nuance is arguably what the user asked for by setting the preference. But it's worth being explicit about, and it interacts with the spinner decision below.

## Consistency question: spinners vs. this dot

The PR notes functional spinners stay unchanged, which is defensible on its own — `.inline-spinner` (`styles.css:1459-1468`, `spinnerSpin ... infinite`) is a case where motion *is* the signal and removing it leaves nothing.

But that's the same argument that applies to `.status-dot-pulse`, which this PR *does* disable, and per the section above the pulse turns out to be load-bearing in two states. So the two calls pull in opposite directions: the highest-motion animation in the app (an unbounded rotation) is exempt as "functional", while a 1.9s opacity fade that carries unique state is not. Either resolution is fine — just worth one line of rationale in the PR so the next person doesn't have to re-derive it.

## `!important` is load-bearing — but only because of where the block sits

It's genuinely required as written, and worth flagging so nobody deletes it on sight:

- `.askcode-loading-pulse` / `.keybinding-recording-pulse` get their animation from an **inline** `style` attribute (`AskCodeCard.tsx:155,168`, `HelpDialog.tsx:417`). Only author-`!important` beats the style attribute.
- The other five base rules live at lines **1274, 1291, 1306, 1325, 1440** — all *after* the media block at **985**, at equal `(0,1,0)` specificity. `@media` adds no specificity, so source order would otherwise win and the override would silently no-op.

**Suggestion:** this file already has a convention that avoids the problem. The repo's four other reduced-motion blocks — `arena-battle.css:121`, `arena-shared.css:232`, `arena-countdown.css:109`, `arena-results.css:346` — all sit at end-of-file and use plain `animation: none` with no `!important`, relying on source order. This block is at 985 of 2211. Moving it to the end of `styles.css` would match that convention and let the five class-based rules drop `!important` (the two inline-style ones still need it).

## Nits

**1. The new test is weaker than it looks.** `reduced-motion-styles.test.ts:36-39` asserts the seven selector substrings and `animation: none !important` **independently** over the block text — not that they're in the *same rule*. I ran the test's exact logic against a deliberately sabotaged stylesheet where the seven selectors got only `transition: none` and the `animation: none !important` sat on an unrelated selector: **it still passes.**

Two smaller things: `css.indexOf(marker)` pins to the *first* block, so one inserted above line 985 would silently redirect every assertion; and it only reads `styles.css`, so the four `src/arena/*.css` blocks aren't covered despite the PR describing this as the shared reduced-motion contract. A single regex pairing the selector list with the declaration would close most of the gap cheaply.

**2. `.task-appearing` now sticks permanently under reduced motion.** With `animation: none` no `animationend` fires, so `TilingLayout.tsx:266-269` never strips the class and it stays on every panel. Harmless while the preference is on — but Chromium re-evaluates `prefers-reduced-motion` live, so toggling the OS setting *off* mid-session makes every mounted panel replay `taskAppear` at once. Gating class *application* on a `matchMedia` signal would avoid it. Edge case; mentioning it because it's cheap to sidestep.

**3. `keybinding-recording-pulse` is applied unconditionally.** `HelpDialog.tsx:367` puts it on every `<kbd>`, but the animation is gated on `recording()` (`:417`, keyframes at `:509`). No bug — `animation: none` on a non-animating element is a no-op — but the name asserts a state that's usually false. `keybinding-key` would read truer.

## Aside — pre-existing, not this PR

`slideUp` (`styles.css:1471`, `translateY(12px) scale(0.97)`) is the one remaining translate-based animation not covered here, which looks like a gap but isn't: it's attached only to `.dialog-overlay > form` and `.dialog-overlay > .dialog-content` (`:1384-1386`), while `Dialog.tsx:87-127` renders `.dialog-overlay` > `.dialog-panel` > `{props.children}`. No `<form>` is ever a *direct* child of the overlay (`NewTaskDialog.tsx:1034` is a grandchild), and `.dialog-content` appears nowhere in the repo outside that selector. Dead CSS — nothing to suppress.

Worth knowing because if it's ever revived by repointing at `.dialog-panel`, it's exactly the motion this PR exists to kill and will need adding to the list.

## Checks run locally

`npm run check` (compile + typecheck + lint + format) clean; the new test passes. All seven selectors verified to resolve to real render sites.

LarryHu0217:
Addressed the review points in `18a214d`:

- Moved the reduced-motion block to the end of `styles.css`; only the two inline-animation overrides retain `!important`.
- Kept the required status pulse suppression, but added a static outer ring so the busy sub-state remains distinct without motion. Functional spinners remain animated where rotation is the loading signal.
- Gated task-entry class application on the current media preference and clear it on `animationcancel`.
- Renamed the unconditional keybinding class to `keybinding-key`.
- Strengthened the test to pair exact selector groups with their declarations, require the block at EOF, and cover the stale-class safeguards.

Validation: `npm run check`, `npm run check:static`, and `npm test` (94 files passed, 2 skipped; 1,573 tests passed, 23 skipped).

johannesjo:
Follow-up review of `18a214d`. I re-checked every point from the last round against the branch — the substantive ones all landed. Two things I'd still fix, plus one approach question that would make both of them moot.

## What I verified as actually fixed

- **Block moved to EOF** (`styles.css:2198`) and `!important` is now confined to exactly the two selectors that need it. That's the correct minimum: author `!important` is the only thing that beats a `style` attribute, so `.askcode-loading-pulse` / `.keybinding-key` genuinely require it and the other five genuinely don't.
- **The strengthened test really is stronger.** I re-ran the sabotage from last round — rewrote the block so the five entry/exit selectors got only `transition: none` while `animation: none` sat on an unrelated selector. The old assertions passed that; the new `expectRule` pairing fails it (`entry/exit group`). All 14 assertions pass on the branch as written.
- **`keybinding-key`** rename ✓ — and suppressing that pulse costs nothing, since recording state is already carried by the `theme.accent` background *and* the "Press shortcut..." label.
- **Status-dot fallback** maps exactly onto `isPulsing()`, so it restores precisely the distinction that was lost. Right call.
- Spot-checked the assumptions underneath it all: `--fg-muted` is defined on `:root` (`styles.css:23`) and every theme override; `onAnimationCancel` is a real Solid prop (`solid-js/types/jsx.d.ts:287`) that falls through to `addEventListener('animationcancel')`; on Electron 40 outlines follow `border-radius`, so the ring is round.

## 1. The status-dot outline leaks onto a non-status element

`.status-dot-pulse` isn't exclusive to `StatusDot`. `TaskStepsSection.tsx:265` (`WaitingIndicator`, rendered at `:391` and `:714`) reuses it purely as an *animation hook* — a 5px `theme.fgSubtle` dot sitting next to the literal text "Waiting for next step".

Under reduced motion that dot now picks up `outline: 1px solid var(--fg-muted); outline-offset: 1px`, which roughly doubles its footprint and encodes a busy sub-state that doesn't exist there and that the adjacent label already spells out — a dimmer dot inside a brighter ring. The comment above the rule ("Preserve a static busy-state distinction") doesn't describe what happens at that call site.

Cheapest fix is to give the ring its own hook so it can't leak — `StatusDot.tsx:52`:

```diff
-      class={isPulsing() ? 'status-dot-pulse' : undefined}
+      class={isPulsing() ? 'status-dot-pulse status-dot-ring' : undefined}
```

…and target `.status-dot-ring` in the media query.

Cosmetic, while you're in there: for `attention` states `getDotShadow` already paints `0 0 0 2px color-mix(…22%)`, occupying 0–2px out from the dot edge. `outline-offset: 1px` puts the gray ring at 1–2px, so it renders *on top of* the colored halo rather than beside it. Worth an eyeball in `needs_input` + busy.

## 2. The stale-entry-class fix stops at the tiling layout

`Sidebar.tsx:1021`, `:1107`, `:1226` apply `task-item-appearing` and never remove it — `TaskRowShell` (`Sidebar.tsx:157-203`) forwards `class` straight to a plain `<div>`, and grepping `src/` for `onAnimationEnd|onAnimationCancel` turns up only the two handlers in `TilingLayout.tsx`. The class is permanently stuck on every sidebar row.

That was harmless before this PR. It isn't now: `.task-item-appearing { animation: none }` means toggling the OS preference *off* mid-session flips `animation-name` from `none` back to `taskItemAppear`, and per spec that starts a fresh animation. So every sidebar row replays opacity 0→1 / max-height 0→40px / translateX −6px→0 simultaneously — the exact scenario `shouldAnimateTaskAppearance()` was added to prevent, at higher volume and without the `animationend` escape hatch the tiling panels have.

Either extend the same gating to `TaskRowShell`, or take the approach below and drop the gating everywhere.

## Approach question: would the near-zero-duration reset be simpler?

Not a blocker, but worth deciding before this lands, because it subsumes both findings above. The standard shape:

```css
@media (prefers-reduced-motion: reduce) {
  *, *::before, *::after {
    animation-duration: 0.01ms !important;
    animation-iteration-count: 1 !important;
    transition-duration: 0.01ms !important;
  }
  /* Rotation *is* the loading signal in icon-only states — no static fallback exists. */
  .inline-spinner {
    animation-duration: 0.75s !important;
    animation-iteration-count: infinite !important;
  }
}
```

Concretely this would delete:

- **the whole `TilingLayout.tsx` change** — at `0.01ms` the animation completes immediately, `animationend` fires, and `task-appearing` strips itself. No stale class, no `animationcancel` handler, no `matchMedia` read, and finding #2 disappears with it;
- **the `!important` reasoning** — a blanket author `!important` already beats the inline `style` attributes on `AskCodeCard.tsx:155,168` and `HelpDialog.tsx:417`, so `.askcode-loading-pulse` / `.keybinding-key` stop being needed as classes at all;
- **most of the new test**, since there's no hand-maintained selector list left to drift.

It would also cover the ~15 `transition:` declarations this PR leaves untouched. Right now we remove a 1.9s opacity fade but keep every transform transition in the app: `button { transition: … transform 0.16s … }` (`:890`) driving `.icon-btn:active { translateY(1px) }` (`:939`), `scale(0.98)` at `:1101,1110`, `translateY(-1px)` at `:1630`, `.focus-mode-task-indicator::before { transition: transform 0.15s }` → `scale(1.15)` (`:1857,1886`), and `transition: rotate 0.14s` on the select picker icon (`:1170`).

I checked the fill-mode hazard and it's clean: none of the pulses use `forwards`, and `statusPulse` / `askcode-pulse` / `keybind-pulse` all start *and* end at `opacity: 1`, so a single instant iteration is a visual no-op. `taskRemove` / `taskItemRemove` do use `forwards`, and snapping to the end state instantly is arguably better than today's `animation: none`, which leaves the panel fully visible for the whole 300ms `REMOVE_ANIMATION_MS` window before it pops out.

The counter-argument is real — the enumerated list is deliberate, avoids blanket `!important`, and matches the four existing `src/arena/*.css` blocks. I'd just rather not maintain a selector list that has no drift protection: a new `@keyframes` added next month is silently uncovered and the new test won't notice. Happy either way, but let's pick one on purpose.

## Nits

1. **`reduced-motion-styles.test.ts:64-70` asserts source text, not behavior.** `expect(tilingLayout).toContain('const appearanceClass = shouldAnimateTaskAppearance()')` breaks on a local-variable rename, and can't detect the class actually being applied under reduced motion. If the gating survives the question above, export `shouldAnimateTaskAppearance` and unit-test it against a stubbed `window.matchMedia` — three cases, including the `typeof window.matchMedia !== 'function'` guard at `TilingLayout.tsx:36`, which is currently untested — then drop the string greps.

2. **The spinner rationale still isn't in the code.** `reduced-motion-styles.test.ts:59` (`expect(block).not.toContain('.inline-spinner')`) is a bare negative assertion, and the explanation lives only in the PR description and the thread. One comment there or next to `.inline-spinner` (`styles.css:1443`) saves the next person re-deriving it.

3. **`expect(css.slice(end + 1).trim()).toBe('')`** turns any future rule appended to `styles.css` into a failure reading `expected '.foo { … }' to be ''`. The intent (source order must beat the base rules) is worth enforcing — just give it a message.

4. **`:53` asserts `outline` but not `outline-offset: 1px`**, which is half the visual fallback.

5. **Test scope vs. name.** It's called `reduced-motion-styles` and framed as the shared contract, but only reads `styles.css` — the four arena blocks (`arena-battle.css:121`, `arena-shared.css:232`, `arena-countdown.css:109`, `arena-results.css:346`) are uncovered. Widen it or rename to something main-stylesheet-specific.

---

Nothing here is hard to land. #1 I'd fix regardless; #2 depends on which way the approach question goes.

LarryHu0217:
Addressed the follow-up in `ca484c6`: isolated the static status ring from the shared pulse hook, gated sidebar entry classes under reduced motion, added behavior tests for the media-query helper, documented the spinner exception, and strengthened the CSS assertions. `npm run check`, `npm run check:static`, and the full test suite pass (1,575 tests; 23 skipped).

johannesjo:
Thank you very much! <3

PR #240, opened 2026-07-23
fix(shortcuts): show feedback for unsupported git actions

LarryHu0217:
## Summary

- show a brief notification when merge or push shortcuts target a non-worktree task
- keep the existing pending-action flow unchanged for worktree tasks
- add focused coverage for both unsupported and supported git actions

## Why

The keyboard handlers queue merge and push actions without checking git isolation. The task panel only opens those dialogs for worktree tasks, so using either shortcut on a direct task previously did nothing and gave no feedback.

This addresses the silent no-op shortcuts item in #214.

## Validation

- `npm test -- src/store/focus.test.ts`
- `npm run check`
- `npm run lint:dead`
- `npm run lint:arch`
- `npm test` (100 files passed, 2 skipped; 1625 tests passed, 23 skipped)

### DIFF
--- a/src/store/focus.test.ts
+++ b/src/store/focus.test.ts
@@ -20,6 +20,7 @@ type MockStore = {
   showPromptInput: boolean;
   sidebarVisible: boolean;
   taskSplitMode: Record<string, boolean>;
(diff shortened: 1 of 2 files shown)


johannesjo:
Thank you very much! <3

PR #241, opened 2026-07-24
fix(ui): use shared font tokens in plan surfaces

LarryHu0217:
## Summary

- use the shared UI font token for plan dialog prose
- use the shared monospace token for plan code and the new-task prompt
- preserve the existing typography sizes and spacing

This addresses the plan-dialog font-stack item in #214.

## Validation

- `npm run check`
- `npm run lint:dead`
- `npm run lint:arch`
- `npm test` (100 files passed, 2 skipped; 1625 tests passed, 23 skipped)


johannesjo:
Thank you very much! <3

PR #246, opened 2026-07-27
Fix keybinding migration notice layout reflow

LarryHu0217:
## Summary
- move the keybinding migration notice outside the main flex layout
- render it as a fixed, responsive notification so showing or dismissing it does not resize task columns and terminals
- preserve the preset and dismiss actions, with accessible labels
- add regression coverage for the fixed-position layout contract

Addresses the keybinding-migration banner reflow item in #214.

## Validation
- `npm test` (101 files passed, 2 skipped; 1,627 tests passed, 23 skipped)
- `npm run check`
- `npm run lint:dead`
- `npm run lint:arch`
- `npm run build:frontend`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -815,75 +815,6 @@ function App() {
             <FocusModeTaskIndicators />
           </div>
         </Show>
-        <Show when={!store.keybindingMigrationDismissed}>
-          <div
-            style={{
-              background: theme.bgInput,
-              border: `1px solid ${theme.border}`,
-              'border-bottom': `1px solid ${theme.border}`,
-              padding: '8px 16px',
-              display: 'flex',
-              'align-items': 'center',
-              'justify-content': 'space-between',
-              'font-size': '13px',
-              color: theme.fg,
-              'flex-shrink': '0',
-            }}
-          >
-            <span>
-              Keyboard shortcuts are now configurable.{' '}
-              <button
-                type="button"
-                onClick={() => {
-                  toggleHelpDialog(true);
-                  dismissMigrationBanner();
-                }}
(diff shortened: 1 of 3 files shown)


johannesjo (commented):
Rechecked the current head against #214, including focus order, overlay layering, responsiveness, and the regression test. The fixed CSS is the right KISS/YAGNI solution; no reusable toast abstraction is needed. CI is green.

Two small, non-blocking follow-ups:

1. `src/App.tsx:863` — the notice is visually at the top but now comes after `<main>` in DOM/tab order. This is not necessarily a WCAG failure, but the move is unnecessary: `position: fixed` already removes the element from flex layout. Keeping the existing `<Show>` before `<main>` preserves the original sequential order and reduces the behavioral change.

2. `src/keybinding-migration-notice.test.ts:10` — the test verifies that the CSS rule exists but never verifies that the notice element uses `keybinding-migration-notice`. Removing or mistyping the class would make the notice a flex item again while every assertion still passes. The smallest fix is a source assertion such as `expect(app).toContain("class=\"keybinding-migration-notice\"")`, ideally scoped to the notice block. If item 1 is adopted, remove or replace the post-`</main>` assertion.

Everything else looks correctly scoped. Ready to merge with only this minor feedback.

LarryHu0217:
Addressed both follow-ups in 5d85c7a: the fixed notice is back before `<main>` in DOM order, and the regression test now verifies the notice class and ordering. Focused tests, `npm run check`, and the full suite (1,627 passed, 23 skipped) are clean.

johannesjo:
Thank you very much @LarryHu0217 ! <3

PR #257, opened 2026-08-09
fix(terminal): increase wheel scroll sensitivity

### LINKED ISSUE #254: Terminal mouse-wheel scrolling is very slow — set xterm's scrollSensitivity
**Environment:** Parallel Code 1.12.0 (deb), Linux (X11, GNOME). Code references verified against `main` @ `d000fff65`.

## Problem

Scrolling back through an agent's terminal with the mouse wheel is slow (working in cc)

## Cause

The `Terminal` in `src/components/TerminalView.tsx` (~L415) is constructed without any scroll options:

```ts
term = new Terminal({
  cursorBlink: true,
  fontSize: initialFontSize,
  fontFamily: getTerminalFontFamily(store.terminalFont),
  theme: activeTerminalTheme(),
  allowProposedApi: true,
  scrollback: TERMINAL_SCROLLBACK_LINES,
  disableStdin: taskPtyDetached(),
  linkHandler: { ... },
});
```

`scrollSensitivity`, `fastScrollSensitivity` and `smoothScrollDuration` don't appear anywhere in the repo, so xterm falls back to its default `scrollSensitivity: 1` — **one line per wheel notch**. Most terminals move 3–5.

## Request

- Set `scrollSensitivity` to something like `3`–`5` by default (and `fastScrollSensitivity` for the Alt-modified fast scroll).
- Better: expose it as a setting, next to the existing terminal font/font-size settings, so people with different mice and trackpads can tune it.
- Optionally `smoothScrollDuration` for smoother motion — though raw speed is the thing that hurts here.

This looks like a one-line default change with a large day-to-day payoff.

## Notes

- #211 proposes the same shape for a different xterm option (a Settings toggle wiring `screenReaderMode` into terminal creation), so this would slot in alongside it.

LarryHu0217:
## Summary
- increase terminal wheel scrolling to four lines per notch
- keep Alt-modified fast scrolling at a five-times multiplier
- apply the shared defaults to desktop and remote terminals
- add focused regression coverage for the scroll options

## Validation
- `npm run test:unit -- src/lib/terminalConstants.test.ts`
- `npm run check`
- `npm run check:static`
- `npm test` (1,805 unit tests and 9 client tests passed)
- `npm run build:frontend`
- `npm run build:remote`

Closes #254


johannesjo:
Thank you very much! <3

PR #258, opened 2026-08-09
feat(review): import GitHub Code Quality findings

### LINKED ISSUE #238: Import GitHub Code Quality findings as an optional provider

LarryHu0217:
Fixes #238

## Summary
- load open GitHub Code Quality findings through read-only `gh api` requests using the user's existing GitHub CLI authentication
- validate and map API responses into the provider-neutral `QualityFinding` review flow
- surface missing CLI/auth, unsupported repository plans, and API failures as non-blocking actionable states
- add an explicit refresh control while keeping automatic loads tied to diff lifecycle changes

## Validation
- `npm run check`
- `npm run check:static`
- `npm test` (1,817 unit tests and 10 client tests passed)
- `npm run build:frontend`
- `npm run build:remote`
- `npm run build:mcp`

### DIFF
--- a/electron/ipc/channel-manifest.json
+++ b/electron/ipc/channel-manifest.json
@@ -112,6 +112,7 @@
   "DetectPrForBranch": "detect_pr_for_branch",
   "RefreshPrChecksWatcher": "refresh_pr_checks_watcher",
   "PrChecksUpdate": "pr_checks_update",
+  "GetGithubCodeQualityFindings": "get_github_code_quality_findings",
   "LogFromRenderer": "log_from_renderer",
   "CheckForUpdates": "check_for_updates",
   "DownloadUpdate": "download_update",
--- a/electron/ipc/github-code-quality.test.ts
+++ b/electron/ipc/github-code-quality.test.ts
@@ -1,0 +1,225 @@
+import { execFile } from 'child_process';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('child_process', () => {
+  const mockExecFile = vi.fn();
+  (mockExecFile as unknown as Record<symbol, unknown>)[Symbol.for('nodejs.util.promisify.custom')] =
+    (file: unknown, args: unknown, opts: unknown): Promise<{ stdout: string; stderr: string }> =>
+      new Promise((resolve, reject) => {
+        mockExecFile(file, args, opts, (error: Error | null, stdout: string, stderr: string) => {
+          if (error) reject(Object.assign(error, { stderr }));
+          else resolve({ stdout, stderr });
+        });
+      });
+  return { execFile: mockExecFile };
+});
+
+import {
+  classifyGithubCodeQualityError,
+  loadGithubCodeQualityFindings,
+  parseGithubCodeQualityFindings,
+} from './github-code-quality.js';
+
+type ExecCallback = (error: Error | null, stdout: string, stderr: string) => void;
+type ExecHandler = (command: string, args: string[], callback: ExecCallback) => void;
+
+function stubExec(handler: ExecHandler): Array<{ command: string; args: string[]; cwd: string }> {
+  const calls: Array<{ command: string; args: string[]; cwd: string }> = [];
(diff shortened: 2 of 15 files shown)


johannesjo:
Thanks for putting this together — and sorry it took a while to get back to you.

I'm going to pass on this one, and the reason is mostly about the underlying GitHub feature rather than your implementation.

**Code Quality is Team/Enterprise-only.** It went GA on 2026-07-20 as a paid add-on for GitHub Team and Enterprise Cloud ($10 per active committer per month, plus usage-based charges). It isn't available on Enterprise Server, and it isn't available on Free/Pro personal accounts. Checking this repo:

```
GET /repos/johannesjo/parallel-code/code-quality/setup
→ 404 {"message": "Code quality is not available for this repository."}
```

So I can't exercise this path end to end, and neither can most people who'd review it. That leaves the error-mapping tests as the only thing standing behind the feature, and they're written against hand-constructed fixtures rather than observed responses. That's a lot of permanent surface for code nobody here can run.

One small correction to the discussion in #238 while I'm at it: `code-quality/findings` returns **404** for this repo, not 403. The 403 (`"You are not authorized to view code quality findings."`) shows up on org-owned repos like `github/docs`. So the "real 403 to test against" doesn't reproduce here.

**Two issues in the diff that would need fixing regardless of the above**, in case they're useful for #260:

1. The provider is on by default for everyone — `findingProvider={devQualityFindingProvider ?? githubCodeQualityFindingProvider}`, and `devQualityFindingProvider` is `undefined` unless a dev-only env var is set. That adds two `gh` subprocess + network round trips (`gh repo view`, then a `--paginate`d `gh api`) to the diff-open path for every user.

2. `HTTP 404` maps to `unavailable(...)`, which `createGithubCodeQualityFindingProvider` rethrows, which sets `findingsError`, which makes `hasReviewSidebarState()` truthy (`ReviewSidebarPanel.tsx:18`). Net effect: the review sidebar force-opens with *"Quality findings unavailable: GitHub Code Quality is not available for this repository."* on every diff open, for every GitHub-hosted repo without a Team/Enterprise plan — including this one. That's the inverse of #238's "personal repositories degrade gracefully" criterion. "Not applicable" and "something went wrong" need to stay distinct.

**Where I'd rather put this effort.** #243 landed the `QualityFinding` contract, but the dev fixture is still the only provider, so the findings half of the review sidebar is dead code in production builds. The first real provider should be one that works for everybody — no auth, no network, no plan gate. `eslint --format json` over the changed files maps nearly 1:1 onto `QualityFinding`, and this repo already runs ESLint 9.

I've opened #260 for exactly that. Given you built the contract in #243, you're the obvious person to pick it up if you're interested — no obligation at all.

Thanks again for the contributions here; the #243 groundwork is good and I want to build on it.

PR #265, opened 2026-08-29
refactor: move bookmark glyph into shared icon

zerone0x:
## Summary
- add `BookmarkIcon` to the shared icon module
- replace the duplicated inline bookmark SVGs in the terminal bookmark gutter
- keeps the two existing rendered sizes (15px and 14px) while removing the local pasted path

Refs #208

## Tests
- `npm ci --ignore-scripts` (plain `npm ci` failed because `node-pty` native postinstall could not rebuild in this container)
- `npm run typecheck`
- `git diff --check`


johannesjo:
Thank you very much! <3

PR #270, opened 2026-09-08
fix(window): make a backgrounded window reachable again

miadisabelle:
"Keep them alive in the background" hides the window instead of closing it, which is the point — the agents keep running. Getting back to it is where it breaks down, in two ways.

**Nothing owns a second launch.** Clicking the icon, or running the binary again, starts a whole second process. It reads the same state file, restores every persisted session, and spawns a duplicate agent for each one — on top of the PTYs the first instance is still holding. The hidden window is not raised, because nothing is listening for the launch. On Linux, where there is no dock to click, that launch is the only gesture a user has, so the app answers "come back" by cloning itself over its own running terminals.

**`show()` is not enough on its own.** It is a no-op on a minimized window, and does nothing for a window that is visible but buried behind another app. Both states reach the same handlers as the hidden one.

### Changes

- Take the single-instance lock in packaged builds, before `fixEnv()` resolves the login-shell environment, and release it at the top of `will-quit` so a replacement spawned by the AppImage updater cannot lose the lock to a process that is already quitting. A second launch now arrives at the running instance as `second-instance` and restores its window instead of starting a process. **Dev runs skip the lock**, so `npm run dev` still starts while an installed build is running.
- Add `restoreWindow()` in `electron/window-restore.ts`: show if hidden, restore if minimized, focus either way, no-op if the window is missing or destroyed. Typed structurally against a `RestorableWindow` shape rather than `BrowserWindow`, so it is testable without an Electron runtime — following the `menu-template.ts` pattern.
- Route all three call sites through it: the existing `activate` handler, the new `second-instance`, and the `before-quit` prompt.

### On that third call site

`before-quit` currently calls `mainWindow.show()` and its comment already says a quit from the menu "must not prompt invisibly". A minimized window is exactly that case, so I included it. Happy to drop it to a separate PR if you'd rather keep this one to the lock.

### Testing

6 unit tests covering hidden, minimized, both at once, visible-but-unfocused, destroyed, and missing. `npm run compile`, `npm run check:static` and the full suite pass.

### Note

Found while running a fork of this app on a headless Linux box, where the icon-relaunch path is the whole story. Not fork-specific — the duplicate-spawn happens on any platform once the window is hidden.

### DIFF
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -82,7 +82,24 @@ function fixEnv(): void {
   }
 }
 
-fixEnv();
+// One running copy per profile, and the lock is taken here rather than beside the
+// window wiring because Electron's guidance is to take it as early as possible and
+// this file gives that guidance teeth: fixEnv() above spawns an interactive login
+// shell, which on a normal rc file (nvm, conda, compinit) costs on the order of half
+// a second. A second launch is going to quit — spending that first would put the
+// delay squarely on the icon-relaunch path the lock exists to make instant.
+//
+// Dev runs skip the lock deliberately, so `npm run dev` still starts while an
+// installed build is running.
+const isPrimaryInstance = !app.isPackaged || app.requestSingleInstanceLock();
+
+if (!isPrimaryInstance) {
+  app.quit();
+} else {
+  // Only the primary instance ever spawns a PTY, so it is the only one that needs
+  // the resolved login-shell environment.
+  fixEnv();
+}
 
 // Blink evicts the oldest WebGL context past 16 per renderer process, and every
 // mounted terminal pane holds one — hidden task/tab terminals included. Past 16
@@ -232,21 +249,14 @@ function createWindow() {
   });
 }
 
-// One running copy per profile. "Keep them alive in the background" hides the
-// window instead of closing it, so a user who launches the app again is asking
-// for the window they already have — but without the lock a second process
-// starts, restores every persisted session from the same state file, and spawns
-// a duplicate agent for each one on top of the PTYs the hidden instance is still
-// holding. The hidden window has no way back either: nothing is listening for
-// the launch. Taking the lock turns a second launch into "show the window".
-//
-// Dev runs skip the lock deliberately, so `npm run dev` still starts while an
-// installed build is running.
-const isPrimaryInstance = !app.isPackaged || app.requestSingleInstanceLock();
-
-if (!isPrimaryInstance) {
-  app.quit();
-} else {
+// Why the lock matters here: "Keep them alive in the background" hides the window
+// instead of closing it, so a user who launches the app again is asking for the
+// window they already have. Without the lock a second process starts, restores
+// every persisted session from the same state file, and spawns a duplicate agent
+// for each one — on top of the PTYs the hidden instance is still holding. The
+// hidden window has no way back either, because nothing is listening for the
+// launch. With the lock, a second launch becomes "show the window".
+if (isPrimaryInstance) {
   // A second launch (icon, CLI, file manager) reaches the instance that owns
   // the lock as this event instead of starting a process of its own.
   app.on('second-instance', () => restoreWindow(mainWindow));
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -91,9 +91,13 @@ function fixEnv(): void {
 //
 // Dev runs skip the lock deliberately, so `npm run dev` still starts while an
 // installed build is running.
-const isPrimaryInstance = !app.isPackaged || app.requestSingleInstanceLock();
+const singleInstanceLockHeld = app.isPackaged && app.requestSingleInstanceLock();
+// Two questions, two names: whether this process holds the lock, and whether it
+// should boot at all. A dev run answers no to the first and yes to the second,
+// which is why one flag covering both would be wrong under either name.
+const shouldStartApp = !app.isPackaged || singleInstanceLockHeld;
 
-if (!isPrimaryInstance) {
+if (!shouldStartApp) {
   app.quit();
 } else {
   // Only the primary instance ever spawns a PTY, so it is the only one that needs
@@ -256,7 +260,7 @@ function createWindow() {
 // for each one — on top of the PTYs the hidden instance is still holding. The
 // hidden window has no way back either, because nothing is listening for the
 // launch. With the lock, a second launch becomes "show the window".
-if (isPrimaryInstance) {
+if (shouldStartApp) {
   // A second launch (icon, CLI, file manager) reaches the instance that owns
   // the lock as this event instead of starting a process of its own.
   app.on('second-instance', () => restoreWindow(mainWindow));
@@ -319,6 +323,19 @@ app.on('before-quit', (event) => {
 // Runs only on a quit that got through the check above, so it cannot destroy
 // anything the user still had a chance to cancel.
 app.on('will-quit', () => {
+  // Hand the lock over before the blocking teardown below, not at process exit.
+  // electron-updater's AppImage path spawns the replacement *before* quitting
+  // (`doInstall` → `spawnLog(destination)`, then `setImmediate(() =>
+  // app.quit())`), so the incoming process is already booting while this one is
+  // still killing agents — and `killAllAgents()` blocks on a `docker kill` per
+  // session. Holding the lock through that can make the replacement fail it and
+  // quit: update applied, app never reappears. Releasing here also covers the
+  // plain case, where someone relaunching during a slow shutdown would
+  // otherwise be handed a window that is already going away.
+  //
+  // `will-quit` only runs on a quit that got past the veto above, so a
+  // cancelled quit correctly keeps the lock. A no-op when none is held.
+  app.releaseSingleInstanceLock();
   killAllAgents();
   // Detached process groups would outlive Electron otherwise.
   verificationRunner.cancelAll();
--- a/electron/main.ts
+++ b/electron/main.ts
@@ -1,5 +1,6 @@
 import { app, autoUpdater, BrowserWindow, Menu, ipcMain, session, shell } from 'electron';
 import { buildMenuTemplate } from './menu-template.js';
+import { restoreWindow } from './window-restore.js';
 import path from 'path';
 import fs from 'fs';
 import { fileURLToPath, pathToFileURL } from 'url';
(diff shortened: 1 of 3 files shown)


johannesjo (commented):
Verified locally: `npm run typecheck`, `npm run check:static` (tsc + eslint + knip + depcruise) and the 7 new unit tests all pass on this branch. The premise holds — before this, nothing listened for `second-instance`, and `activate` called only `show()`. The `if (!gotTheLock) quit() else { on('second-instance'); whenReady() }` shape matches Electron's canonical example, and adding `show()` on top of the documented `isMinimized() → restore(); focus()` is the right extension for an app that actually hides its window.

Four things, one of which I'd fix before merge.

### 1. `fixEnv()` runs before the lock is requested — `main.ts:84` vs `main.ts:240`

The second, doomed process still executes `fixEnv()` at module scope: an interactive+login shell (`-ilc`) that sources `.zshrc`, so nvm/conda/compinit all run before it reaches line 240 and finds out it should quit. That's dead time sitting directly in the path this PR exists to fix — the user clicks the icon and nothing happens until a throwaway shell resolves.

I timed the exact probe from `fixEnv()` on my machine: **~0.53s**, bounded at 5s by the existing `timeout`. Not fatal, but it's a half-second of "did my click register?" on the icon-relaunch path, which per your Note is the whole story on the headless Linux box.

Electron's guidance is to take the lock as early as possible. Hoisting the check above `fixEnv()` and moving `fixEnv()` into the primary branch would fix it — the resolved environment only matters for spawning PTYs, which only the primary does.

### 2. The visible-but-buried case may not be deliverable on Wayland (note, not a change request)

Electron's own docs for `win.focus()`: *"On Wayland (Linux), the desktop environment may show a notification or flash the app icon if the window or app is not already focused."* So of the three states the PR enumerates, the third one — visible but behind another app — may resolve to an icon flash rather than a raise on Wayland, since the compositor generally won't let a client raise itself.

I don't think there's a code fix here (`show()` isn't demonstrably better under Wayland, and deviating from the canonical pattern to chase it isn't worth it). But the `restoreWindow` doc comment states the buried case as solved, and on one of the two Linux display servers it likely isn't. Worth a sentence in the comment so the next person debugging "focus doesn't work on my desktop" doesn't go looking for a bug in this function.

I'd also drop the "show() is a no-op on a minimized window" claim in the header comment — Electron's docs don't actually say that either way, and the test file's comment (*"minimized windows report themselves as not visible on some platforms"*) tells a different story from the source comment. The code calls both and is correct regardless; only the narration is over-specified.

### 3. `restoreWindow`'s doc comment names a `tray` entry point that doesn't exist — `window-restore.ts:8`

No Electron `Tray` anywhere in the repo (the only hit is a renderer component, `NeedsInputTray`). Either drop it or the next reader goes hunting for a tray handler.

### 4. The dev exemption sanctions a config that cross-wires agent hooks (pre-existing; probably its own PR)

Not introduced here, but the new comment promotes "dev alongside an installed build" to a documented, supported configuration, so it's worth knowing what that configuration does today: `persistence.ts:6-12` gives dev its own `-dev` state dir, but `agent-hooks/runtime.ts:18` uses raw `userData`. Both instances write `agent-hooks/endpoint` (port + token) to the same path, and the hook script re-sources that file on every event (per `hook-script.test.ts:20`) — so whichever instance started last receives hook status events for *both* instances' agents. A one-line fix would mirror the dev suffix from `persistence.ts`.

### Minor

- `second-instance` drops `commandLine`/`workingDirectory`. Correct today — `main.ts` reads no argv (the only `process.argv` is in `electron/mcp/server.ts`, a separate process). Flagging it because `second-instance` is also where deep links land on Linux/Windows, so this is the hook to remember if a `parallel-code <dir>` or protocol handler ever ships.
- `'asks whether the window is destroyed before touching it'` asserts a mock was called, which the no-op test already covers. Harmless, adds nothing.

### On your question about the third call site

Keep it here. One line, same failure mode, and the existing comment already asserted a property the code didn't have — splitting it would leave that comment lying for another cycle.

### Caveat on coverage

The lock only engages in packaged builds, so neither CI nor `npm run dev` exercises the `second-instance` path at all — the unit tests cover `restoreWindow` in isolation, not the wiring. Worth one manual packaged-build check (launch, hide, relaunch) before merge.

miadisabelle:
Thanks — the `fixEnv()` catch is the good one. Pushed 64123b3.

## 1. Lock before `fixEnv()` — fixed

The lock decision now happens before `fixEnv()`, and `fixEnv()` moved into the primary branch on your reasoning: only the primary spawns PTYs, so only the primary needs the resolved environment.

I measured it on a packaged build rather than trusting the shape of the fix. Xvfb + kwin, a throwaway `--user-data-dir`, 3 runs each, timing the doomed second instance end to end:

| | second instance | login-shell probes observed while it ran |
|---|---|---|
| before | **7.86s** (7.80 / 7.75 / 8.04) | 63 |
| after | **1.25s** (1.24 / 1.23 / 1.27) | 0 |

Your 0.53s is the floor, not the ceiling. This box is the headless Linux one from the Note, and its rc file has nvm and conda in it — so the probe costs 6.6s here, and every bit of it landed on the click.

## 2. Wayland — noted in the comment, no code change

Agreed on both halves. The doc comment now says the visible-but-buried case may end at an icon flash rather than a raise on Wayland, with the reason (a client generally cannot raise itself; Electron's docs say `focus()` may notify or flash instead), and that hidden and minimized are unaffected.

I also dropped the "`show()` is a no-op on a minimized window" claim. You were right that the source comment and the test comment were telling different stories, and the docs back neither. The comment now says only what the code relies on: the two states are not exclusive, so both questions get asked and each answer gets acted on.

## 3. `tray` — dropped

No Electron `Tray` in the repo. Gone.

## 4. `agent-hooks/endpoint` cross-wiring — separate PR, ready when you want it

Confirmed the shape you describe: `persistence.ts:6-12` gives dev its own `-dev` state dir, `agent-hooks/runtime.ts:18` uses raw `userData`, and the hook script re-sources the endpoint file per event — so the last instance to start collects hook status for both. Happy to send the one-line mirror of the dev suffix as its own PR, since as you say it predates this one. Say the word and it's up.

## Minor

- Dropped the `isDestroyed()` mock assertion. You're right, the no-op test already covers it.
- `commandLine`/`workingDirectory`: understood, leaving the parameters off rather than accepting and ignoring them. Noted that this is where deep links would land.

## Third call site

Kept, per your call.

## Packaged-build check

Done, since neither CI nor `npm run dev` reaches this path. Packaged build, Xvfb + kwin, both instances on the same `--user-data-dir`:

```
A launched            → window state: Normal, 2 processes on the profile
xdotool windowminimize → window state: Iconic
B launched            → exit 0 after 1.14s
A still running       → YES
A window state        → Normal        (restored by second-instance)
processes on profile  → 2 (unchanged)
```

So the wiring is exercised, not just `restoreWindow` in isolation: B declines to become a second app, and the minimized window comes back. I used minimize rather than "Keep in Background" because driving the UI headless is awkward, and minimize is the stricter of the two — it needs `restore()`, where hide only needs `show()`.

`npm run check:static` and the full suite (2219 passing) are green on the branch.

johannesjo:
Re-reviewed at `64123b3`. `npm run typecheck`, `npm run check:static` (tsc + eslint + knip + depcruise, 507 modules / 1650 deps) and the `window-restore` suite all pass locally.

All four points from the last round are closed:

1. **Lock before `fixEnv()`** — fixed at `main.ts:93`, `fixEnv()` moved into the primary branch. Your 7.86s → 1.25s is the same effect I measured at 0.53s, scaled to an rc file with nvm and conda in it. Measuring it on a packaged build instead of trusting the shape of the fix is the right way to have closed it.
2. **Wayland** — caveat is in the header at `window-restore.ts:11-14`, and the over-specified `show()`-on-minimized claim is gone. Agreed on both.
3. **`tray`** — gone.
4. **agent-hooks cross-wiring** — agreed, its own PR. Send it whenever; it predates this one.

I also checked what the doomed second instance still does before it quits, since module scope keeps running past `app.quit()`: `will-quit` fires, but `killAllAgents`, `cancelAll`, `stopAllPlanWatchers`/`stopAllStepsWatchers` are all empty-state no-ops there, and `stopAgentHookRuntime()` returns early on a null server — nothing in `agent-hooks/` unlinks the endpoint file, so the doomed instance cannot damage the primary's shared state on its way out. That was the thing I most wanted to be wrong about, and it's clean.

One new thing, which I'd like your read on before merge.

---

### The lock is never released on the way out, and the AppImage updater relaunches before this process exits

`electron-updater` defaults `autoRunAppAfterInstall = true` (`AppUpdater.js:119`), so `quitAndInstall()` → `install(false, true)` → `AppImageUpdater.doInstall` calls `spawnLog(destination)` — **spawning the replacement AppImage** — and only then `setImmediate(() => { emit('before-quit-for-update'); app.quit(); })` (`BaseUpdater.js:12-22`). So the new process is already booting while this one still holds the lock, and if it reaches `main.ts:93` first it fails the lock and quits: update applied, app never reappears.

Whether it reaches it first is a race, so I measured both sides on this box rather than guess:

| | |
|---|---|
| replacement: spawn → `requestSingleInstanceLock()`, real module graph | **236–348 ms** (5 runs, unpacked `dist-electron`, warm cache) |
| outgoing: `docker kill` per session in `killAllAgents` (`pty.ts:671-688`), serial | **114–129 ms** (3 runs) |

An AppImage adds squashfs mount on top of that first number, which *helps* here — a slower replacement is a safer one. So this only bites when teardown outlasts the replacement's boot, which at these rates means roughly ten-plus docker-backed sessions, or fewer with a slow daemon (each kill is capped at the existing `timeout: 3000`, not 120 ms). I should correct myself on one thing: my first read of this was that a couple of containers would stall teardown for seconds. That isn't what the numbers say — 120 ms each is the normal case, and the 3 s is a ceiling for a hung daemon, not a typical cost.

So: narrower than I first thought, but the crossover is squarely inside what this app is for — parallel docker agents are the premise, and ten of them is a normal afternoon. I have **not** reproduced it end to end; that needs a packaged AppImage and a real update, which I can't drive here.

Deb builds are unaffected (`isAutoUpdateSupported` requires `process.env.APPIMAGE`), and so is macOS — `MacUpdater.handleUpdateDownloaded` hands off to Squirrel's native `quitAndInstall`, which relaunches after the process exits, not before.

The fix is one line, deterministic, and doesn't depend on which side usually wins — release the lock as soon as a quit is committed, ahead of the blocking kills:

```ts
app.on('will-quit', () => {
  // The AppImage updater spawns the replacement before this process exits, and
  // killAllAgents() below can block on `docker kill` per session. Hand the lock
  // over now so the incoming instance does not lose the race and quit.
  app.releaseSingleInstanceLock();
  killAllAgents();
  ...
```

`will-quit` rather than the `before-quit-for-update` listener because it also covers the plain case: a user who relaunches during a slow shutdown currently reaches a dying instance via `second-instance` and gets `restoreWindow` on a window that is already going away. And `will-quit` only runs on a quit that got past the veto at `main.ts:305`, so a cancelled quit correctly keeps the lock. It's a no-op in the doomed second instance, which never held it.

---

### Minor

- `isPrimaryInstance` (`main.ts:93`) is `true` for *every* dev instance, so the name promises a uniqueness it doesn't have in dev. `ownsSingleInstance` reads straighter. The comment explains it, so take it or leave it.
- PR description still says "7 unit tests"; there are 6 since the mock assertion came out. Worth a touch-up before it becomes the commit message.

### Verdict

Needs one line in `will-quit`, and I'd like your take on the AppImage argument above in case you think the race is too narrow to be worth pre-empting. Everything else is ready — the structural `RestorableWindow` typing is what makes the tests possible without an Electron runtime, the test comments say *why* each state needs its own call, and the packaged-build check I asked for is done.

miadisabelle:
Right on the ordering — `doInstall` → `spawnLog(destination)` → `return true`, then `setImmediate(… app.quit())`. Taken, plus a rebase onto 2.0.0. Head is `b02a018`.

## The line

Released at the top of `will-quit`, with your reasoning in the comment. One refinement to the "plain case" argument: by the time `will-quit` runs, every window is destroyed and `mainWindow` is null, so a relaunch during a slow shutdown reaches `second-instance` and `restoreWindow(null)` returns immediately. It does nothing at all rather than surfacing a dying window — which is still the thing worth fixing, just for a duller reason than either of us wrote down.

## Your race, measured on this box

You measured on yours; here is the other data point. Same two quantities, headless Linux:

| | your machine | this box |
|---|---|---|
| incoming: spawn → `requestSingleInstanceLock()` | 236–348 ms | **742–793 ms** (packaged `linux-unpacked`, 5 runs) |
| outgoing: serial `docker kill` per session | 114–129 ms | **263–306 ms** (1/5/10 containers) |

Both sides are slower here, but not by the same factor, and the crossover moves the wrong way: teardown passes the incoming instance's 770 ms at **three** docker-mode sessions, not ten. An AppImage's squashfs mount pushes the incoming number up and helps, as you said — but the direction of the correction is that slower hardware widens the affected population rather than narrowing it. Your "narrower than I first thought" reads to me as still too generous to the current code.

## What I could not show

I tried to observe the race end to end rather than argue it, and failed. I built a packaged harness with a switchable release and a stand-in block in `will-quit`, plus a separate Electron probe that calls `requestSingleInstanceLock()` and prints the result. The control is sound — the probe gets `DENIED` against a healthy instance and they share the same `SingletonLock` — but the probe's own boot is ~4 s via `npx electron`, which swamps a 4 s window completely. Both arms returned `ACQUIRED` within noise. That experiment distinguishes nothing, so I'm reporting it as inconclusive rather than dressing it up. The code-path argument and the two measurements above are what I'd stand behind.

## A residual the release introduces, bounded

Worth your call rather than mine. Releasing early lets the incoming instance start while the outgoing one is still killing containers, and container names come from a persisted id — `parallel-code-${agentId.slice(0, 12)}` at `pty.ts:313`. Restored agents carry `attachExisting: true`, but a fresh process has no `sessions` entry, so it spawns; there is no `docker rm -f` first, and Docker rejects a duplicate name outright rather than waiting:

```
docker: Error response from daemon: Conflict. The container name
"/parallel-code-abc123def456" is already in use by container "4987f63d3396…".
```

So the incoming instance's restored agent fails to start rather than being silently killed. Reachability: the incoming instance needs to reach its first container spawn before the outgoing one has killed that container, and `docker kill` frees the name in ~322–475 ms here. Against a realistic boot to first spawn of 3–5 s — `fixEnv`, `startAgentHookRuntime`, bundle, `loadState`, plus `RESTORED_AGENT_SPAWN_STAGGER_MS` of 1000 per extra agent — that needs roughly **10–16 docker-mode agents restored at once**, and both sides iterate in about the same order so the first spawn races the first kill and loses by seconds.

Narrower than the failure the line fixes, and the alternative fails 100% of the time in the case the line exists for. I'd take the line and leave this; flagging it so it isn't a surprise later.

## `isPrimaryInstance`

Took the complaint, not quite the name. `ownsSingleInstance` has the same problem it's meant to solve — a dev run owns no lock either, and would still be reading `true`. Split into two flags instead, each true only where it reads as true:

```ts
const singleInstanceLockHeld = app.isPackaged && app.requestSingleInstanceLock();
const shouldStartApp = !app.isPackaged || singleInstanceLockHeld;
```

Same short-circuit as before: `requestSingleInstanceLock()` is called exactly once when packaged and never in dev. If you'd rather have one name, say which and I'll collapse it.

## Test count

Fixed in the PR description, which is the one that becomes the squash message — it now says 6. I left the `f42175d` body at 7, since that commit's own tree has 7 and the next commit removes the seventh and says so.

## Checks

Rebased onto `5b0ba39` (2.0.0). `npm run compile`, `npm run check:static` (646 modules, no violations), `vitest run` 2541 passed, `test:client` 244 passed.

One thing from #272 that also applies here: `npm run typecheck` and `npm run check:static` don't type-check `electron/**` — root `tsconfig.json` has `"include": ["src", "electron/ipc/channels.ts"]`. Only `npm run compile` covers the main process, and this PR is all main-process code.

johannesjo:
Thank you very much! <3

PR #272, opened 2026-09-08
fix(agent-hooks): keep a dev run's endpoint file out of the installed profile

miadisabelle:
The dev-instance cross-wiring you spotted reviewing #270. Your diagnosis was right; two things about it turned out to be worse than the comment said, and the fix wanted to be slightly bigger than one line.

### Why it isn't one line

The suffix rule already exists twice, verbatim in shape: `getStateDir()` in `persistence.ts:5` and `getKeybindingsDir()` in `register.ts:769`. It is missing from three places — `agent-hooks/runtime.ts:18` and both arena handlers. Mirroring it into `runtime.ts` would have made a third copy of a rule that has now been forgotten more often than it has been remembered.

So `electron/user-data-dir.ts` instead: `resolveUserDataDir(userDataPath, isPackaged)` is pure, so the rule is testable without an Electron runtime, and `getUserDataDir()` is what callers use. After this, `app.getPath('userData')` appears exactly once in the tree, which makes "someone reached past the helper" a one-line grep instead of an audit.

Same reasoning you used on #270 for the third call site: leaving the copies in place would leave the next one to drift.

### Worse than the comment said

**The misroute succeeds.** The port *and* the token both come from `endpoint.env`, so an agent launched by instance A posts to instance B and passes `tokenMatches` cleanly. Nothing logs on either side — `coordinator.ts:318` returns early on an unknown agent id, and `applyAgentHookEvent` never checks the agent belongs to this instance, so the wrong instance writes phantom status entries.

**Quitting one instance breaks hooks for the other.** `endpoint.env` is written only inside `startAgentHookServer`, at startup. Close the instance that owns the file and it still names a dead port, so the surviving instance's agents post into nothing and fall back to PTY heuristics until that instance is itself restarted.

### Observed, not just reasoned

Neither CI nor the tests exercise the dev/packaged separation — the unit tests cover the string rule only. Same gap as the lock in #270, so the same treatment: packaged build under Xvfb, dev run started second, both pointed at the same `--user-data-dir`.

**Before** (upstream `runtime.ts`) — the dev run overwrites the packaged instance's file:

```
packaged run  → prof/agent-hooks/endpoint.env      PORT=37127
                prof-dev/agent-hooks/endpoint.env  (absent)
dev run       → prof/agent-hooks/endpoint.env      PORT=41559   ← overwritten
                prof-dev/agent-hooks/endpoint.env  (absent)
```

**After** — the packaged file is untouched:

```
packaged run  → prof/agent-hooks/endpoint.env      PORT=42473
                prof-dev/agent-hooks/endpoint.env  (absent)
dev run       → prof/agent-hooks/endpoint.env      PORT=42473   ← unchanged
                prof-dev/agent-hooks/endpoint.env  PORT=46743
```

### Packaged builds are unchanged

`resolveUserDataDir` returns its input on the first line when `isPackaged`, and `getUserDataDir()` is its only caller. All four converted sites resolve to exactly the same paths they did before in a packaged build. agent-hooks needs no migration either — `endpoint.env`, `hook.sh` and `claude-settings.json` are rewritten every launch.

### The second commit is droppable

`SaveArenaData`/`LoadArenaData` are the same slip without the silent misrouting, found while making the first fix. It is a behavior change for dev runs: `arena-presets.json` and `arena-history.json` stay with the installed build, so a dev run picks this up as an empty arena. Split out so you can drop it and keep this PR to agent-hooks.

### Checks

`npm run compile`, `npm run check:static`, `vitest run` (2216 passed / 25 skipped) and `npm run test:client` (36 passed / 6 skipped) all green.

One note in case it is useful to you beyond this PR: `npm run typecheck` and `npm run check:static` do not type-check `electron/**` — root `tsconfig.json` has `"include": ["src", "electron/ipc/channels.ts"]`, so only `npm run compile` covers the main process. I had a main-process type error passing both of the checks you used to verify #270 before I caught it.

### DIFF
--- a/electron/agent-hooks/runtime.ts
+++ b/electron/agent-hooks/runtime.ts
@@ -1,6 +1,7 @@
-import { app, type BrowserWindow } from 'electron';
+import { type BrowserWindow } from 'electron';
 import path from 'path';
 import { IPC } from '../ipc/channels.js';
+import { getUserDataDir } from '../user-data-dir.js';
 import { setAgentHookRuntime } from '../ipc/pty.js';
 import { error as logError, info as logInfo } from '../log.js';
 import { emitAgentHookEvent } from './events.js';
@@ -15,7 +16,11 @@ let server: AgentHookServer | null = null;
  * Failure is logged and otherwise ignored: the PTY heuristics keep working.
  */
 export async function startAgentHookRuntime(getWindow: () => BrowserWindow | null): Promise<void> {
-  const dir = path.join(app.getPath('userData'), 'agent-hooks');
+  // Per-instance, not raw userData: this directory holds endpoint.env (loopback
+  // port + bearer token), written only at startup, so sharing it between a dev
+  // run and an installed build hands every agent's hook events to whichever
+  // instance started last.
+  const dir = path.join(getUserDataDir(), 'agent-hooks');
   // Resolved per event: the server starts before the window exists so that no
   // Claude launch can race it, and the window may be recreated later.
   const forward = (event: AgentHookEventPayload): void => {
--- a/electron/ipc/persistence.ts
+++ b/electron/ipc/persistence.ts
@@ -1,19 +1,9 @@
-import { app } from 'electron';
 import fs from 'fs';
 import path from 'path';
-
-function getStateDir(): string {
-  let dir = app.getPath('userData');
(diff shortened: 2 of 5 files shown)


johannesjo:

PR #275, opened 2026-09-12
fix(permissions): resolve skip-permissions flags at every launch path

miadisabelle:
Four call sites decide an agent's skip-permissions flags by reading `skip_permissions_args` off whatever `AgentDef` they happen to hold. When a def arrives carrying none, an explicit `skipPermissions: true` produces a launch that prompts on every tool call — nothing logs, and the setting reads as broken.

`enrichAgentDef` already repairs one shape of this on restore, so the gaps it leaves are the interesting part: it tests `!agentDef.skip_permissions_args`, so a def carrying `[]` is never repaired, and it matches on `availableAgents.find(a => a.id === agentDef.id)`, so the `MCP_TaskCreated` fallback def — whose id is the bare command, never `claude-code` — and every `custom-*` agent are never repaired at all.

### Why it needs a shared module

`getSkipPermissionsArgs` already exists in `electron/ipc/agents.ts`, but only the main process can call it: that file imports `child_process` for the `which` probe, and the renderer-imports-main rule correctly forbids reaching it. The renderer builds launch args synchronously at spawn time with no IPC round-trip available, so it has no way to ask the same question.

The first commit moves the flags to `electron/shared/skip-permissions.ts`, which is already on the allowlist for renderer-safe modules. No new architecture rule, no change to `.dependency-cruiser.cjs`. `DEFAULT_AGENTS` now takes its `skip_permissions_args` from that table instead of repeating the literals, and the coordinator's import moves with the function.

That commit is behaviour-neutral. I compared the old and new implementations over 27 inputs — the six built-ins, absolute and relative paths, trailing slashes, `''`, `/`, `//`, a Windows-style path, unknown commands, and every inherited object key — and they agree on all of them. Two details that make that true rather than nearly true: empty path segments are dropped so a trailing slash resolves the way `path.basename` did, and the table is a `Map` rather than an object literal, because `command` is free text from the custom agent editor and an object lookup would resolve `constructor` or `toString` to a prototype member instead of returning nothing.

### The fixes

| | |
|---|---|
| `buildTaskAgentArgs` | emitted nothing when the def's list was empty, so a degraded def silently downgraded an opt-in to a normal launch |
| `MCP_TaskCreated` fallback def | builds `skip_permissions_args: []` when `availableAgents` has no entry for the coordinator's command. That def is persisted under `id: cmd`, which `enrichAgentDef` never matches, so the empty list is stored permanently and read by every later consumer |
| New Task dialog | hid the checkbox for a def with no flags rather than resolving by command, so an agent restored from an older profile offered no way to opt in and recorded `skipPermissions: false` for the life of the task |
| `handleCreateTask` (remote) | never passed `skipPermissions` at all, so every task created from a paired phone launched bare regardless of the setting |

The first three ask `resolveSkipPermissionsArgs`, which prefers what the def carries and falls back to the table by command basename. A def with its own flags, including a custom one, is unaffected.

### One behaviour change worth your call

A custom agent whose command basename is one of the built-ins and whose skip-args field was left blank now resolves to the built-in flag, where before a blank field meant none. That follows directly from the "carries nothing, so ask the command" rule the whole fix rests on, but it lands on user-authored config rather than only on degraded state, so it should be your decision rather than a side effect you find later. If you'd rather keep blank-means-none for `custom-*` agents specifically, that is a one-line guard in `resolveSkipPermissionsArgs` and I'll add it.

### Defaults

None change. `defaultSkipPermissions` still defaults to false, and a task still gets the flag only where it opted in. The remote commit only propagates a choice already made on the desktop to a surface that was ignoring it.

### Split

Three commits, each standing alone — I ran `compile`, `check:static`, `lint:dead`, `lint:arch` and the full suite at each one:

1. the move — behaviour-neutral, takeable by itself
2. the three degraded-definition paths
3. the remote path — a separate bug in a separate file, droppable without touching the other two

`electron/ipc/agents.test.ts` held exactly one test, for the function that moved; it moves to `electron/shared/skip-permissions.test.ts` alongside its subject, which leaves that file empty and removed.

### Checks

`npm run compile`, `npm run check:static` (645 modules, no violations), `npm run lint:dead`, `vitest run` 2555 passed, `test:client` 244 passed. 20 new tests.


johannesjo:
Thank you very much fro this! <3

PR #277, opened 2026-09-14
fix: recognize Codex CLI "Ask Codex to do anything" ready prompt

Deltoyd:
## Summary

Parallel Code 2.0.0 cannot deliver automatic sub-task prompts to Codex CLI v0.154.0 when launched through a custom `codex-nono` agent. The child process reaches:

```
> Ask Codex to do anything
```

but the task stays stuck at "Waiting to send prompt..." — the initial prompt and any queued follow-up never send.

`AGENT_READY_TAIL_PATTERNS` in `electron/shared/prompt-detect.ts` only recognized `> Type your message` (Gemini) or a bare `>` as a ready main-input prompt for the `>`-style agents. Codex CLI 0.154.0's placeholder text (`Ask Codex to do anything`) fell through, so `getAgentPromptReadiness`/`chunkContainsAgentPrompt` never reported the agent as ready and the coordinator never wrote the queued prompt.

- Extends the `>`-prompt pattern to also match `Ask Codex to do anything`
- Adds a recorded fixture + targeted unit tests reproducing the reported frame
- Fixes `coordinator-test-harness.ts`'s hand-duplicated copy of the same regex (used to mock `../shared/prompt-detect.js` in `coordinator.test.ts`), which was out of sync with the real pattern and caused the new fixture's `it.each` coordinator tests to fail even after the real fix

## Test plan

- [x] `npx vitest run electron/mcp/prompt-detect.test.ts` — 58/58 pass (new tests fail without the fix, confirmed via reverting locally)
- [x] `npx vitest run electron/mcp/coordinator.test.ts` — 285/285 pass
- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npm run test:unit` full suite — no new failures introduced (compared against a clean checkout of `main` via `git stash`); 8 pre-existing failures unrelated to this change remain (local git identity requirements in `documents/setup.test.ts`, a keychain-fallback token leak in `claude-usage.test.ts`, and a `recovered` flag regression in `plans.test.ts` — none touched by this diff)

https://claude.ai/code/session_0159pgngC1aN4DEVBQoiCGNc


Deltoyd:
Update: my first pass matched the wrong prompt character. The bug report transcribed the prompt as literal `> Ask Codex to do anything`, but a screenshot of the actual app shows Codex still renders its usual `›` character, just with the new placeholder text trailing it (`›  Ask Codex to do anything`) instead of standing alone. Pushed a follow-up commit that merges the bare-`›` and `>`-with-placeholder patterns into one that accepts either prompt character with either placeholder (or a bare prompt, preserving prior behavior), and corrected the fixture/tests to use the real `›` character. Verified directly against the literal text from the screenshot.

PR #280, opened 2026-09-19
test: keep the client suite independent of the host locale

miadisabelle:
Two client tests assert English output from code that formats in the host's default locale — correct for the app, but it fails the suite on a developer machine set to another language.

`npm run test:client` on a clean checkout of `main`:

| `LC_ALL` | before | after |
|---|---|---|
| `fr_FR.UTF-8` | 3 failed | 0 |
| `de_DE.UTF-8` | 4 failed | 0 |
| empty (Intl resolves `und`) | 2 failed | 0 |
| `en_US.UTF-8` | 0 | 0 |

### SessionPicker

`relativeTime` formats with `Intl.RelativeTimeFormat(undefined, …)` and its tests match `/day/` and `/second/`. Two days ago is `'avant-hier'` under fr_FR, `'vorgestern'` under de_DE, and `'-2 d'` under `und`.

The function already takes `now` so its tests are deterministic. It now takes an optional `locale` for the same reason, and the tests pass `'en'`. The call site passes nothing, so the app still formats in the user's locale.

### AgentChatView

The context and token titles use `toLocaleString()`, and the tests expected `'50,000 of 200,000'` and `'12,500 tokens'` — `'50.000'` under de_DE, U+202F-grouped under fr_FR. The expected text is now built with `toLocaleString()`, as `AnnotationMarker`'s test already does.

If you'd prefer one line instead, setting `test.env.LC_ALL` in `vitest.client.config.ts` also works, since the forks pool starts its workers with that environment — but it pins every client test to English and depends on the pool staying process-based, so I went with fixing the two tests.

`npm run compile`, `typecheck`, `lint` pass.


miadisabelle:
Rebased onto 3.0.1 (`f190ba0`), no conflicts. Re-measured `npm run test:client` on the rebased branch: 930 passed under `LC_ALL=fr_FR.UTF-8`, `de_DE.UTF-8`, empty and `en_US.UTF-8` — none of the client tests added since assume English.

PR #282, opened 2026-09-19
fix(ui): hand focus to a column clicked while the sidebar has it

miadisabelle:
After moving through the sidebar with Alt+↑/↓, clicking a task or terminal column's title bar or header selects it but leaves the keyboard with the sidebar: no panel shows focus, typing goes nowhere, and Alt+↑/↓ keeps moving the sidebar highlight.

### Why

`sidebarFocused` is a hard gate — while it's set, `isPanelFocused` returns false for every panel and `scheduleTaskFocus` declines to move DOM focus. The column's pointer handlers call `setActiveTask`, which clears the new-task panel and placeholder gates but deliberately not this one, since keyboard jumps share it and are meant to keep the sidebar focused. The title bar's drag handler also prevents the mousedown default, so a tap there moves no DOM focus on its own.

Clicks *inside* a terminal, notes or prompt were already fine — those handlers call `setTaskFocusedPanel`, which clears the gate.

### Fix

`activateTaskFromPointer` activates the task and clears `sidebarFocused` when `setActiveTask` accepted the id. For a column that was **already** active while the sidebar had focus, nothing about the selection changes, so its focus effect doesn't re-run — the function moves DOM focus to `getTaskFocusedPanel(id)` itself, the same target Enter on a sidebar row uses.

The pointer-driven switchers use it: a task or terminal column, a title-bar tap, a sub-task chip, a focus-mode pill. Sidebar clicks, keyboard jumps and in-column handlers keep `setActiveTask`.

**Out of scope:** the document workspace column activates through its own focus-preserving handler (`preserveInteractionFocus`), so I left it alone rather than route it without understanding that logic.

### Tests

- **4 client tests** on the real store, focus registry, `focusSidebar`/`navigateRow` and the title bar's `handleDragReorder`, with a copy of TaskPanel's focus effect (mounting `TaskPanel` itself would pull in most of the app). A control shows `setActiveTask` leaving DOM focus on the sidebar; then focus lands in another column, back in the already-active one (fails without that branch), and stays where it was when the sidebar didn't have it (fails without the `leavingSidebar` guard).
- **5 store tests** against the real focus gate; the two covering the fix fail without its clearing line.
- `TaskPanel.reasoning.client.test.tsx` mocks the store module explicitly, so it gains the new function beside `setActiveTask`.

`npm run compile`, `check:static` and both suites pass.


miadisabelle:
Rebased onto 3.0.1 (`f190ba0`). The only conflict was `SubTaskStrip.tsx`, where 3.x wraps the chip in a new `<span>`; the change there is still the one call. The bug is unchanged on 3.x (`setActiveTask` still leaves `sidebarFocused` set, and `scheduleTaskFocus` still declines while it is). 3.x's `scheduleTaskFocus` now reads `document` in its microtask, which the Node store tests don't have, so those tests stub it and assert that focus is requested; DOM focus stays covered by the client tests. `compile`, `check:static`, 3545 unit and 934 client tests pass.

PR #283, opened 2026-09-19
feat(tasks): let skip-permissions be changed on an existing task

miadisabelle:
Whether a task's agent launches with its skip-permissions flag is decided when the task is created and fixed for its whole life. Changing it means closing the task and starting again, which loses its worktree and conversation — and the title bar doesn't show which tasks are running with the flag.

### What changes

- **`setTaskSkipPermissions(taskId, enabled)`** changes the flag on an existing task and saves. It takes effect the next time the agent starts — a terminal restart or resume, or a new or reconnected chat. A running agent keeps the permissions it was launched with; I traced every restart path (`restartAgent`/`switchAgent` bump `generation`, which remounts `TerminalView` and rebuilds args) to confirm the flag is read at spawn, not cached.
- **A clickable badge** in the title bar, after the verification badge and styled like the landing badge: "skip confirms" in the warning colour when on, "confirms on" muted when off. It appears when one of the task's agents takes such a flag (resolved by command via `resolveSkipPermissionsArgs`, so a definition restored without its flags still qualifies) and not on landed tasks.
- **The chat view's permission selector** now follows the running session's own mode instead of the task flag. The session refuses a mode change based on how it was started (`claude.ts` — "This task skips permissions; turn that off to choose a mode."), and once the flag can change mid-chat, reading it would enable the selector for a session that rejects the choice. Claude reports its mode only in each turn's `init` message, so a session launched skipping permissions now sets `bypassPermissions` itself on connect — otherwise the selector would be enabled from connect until the first prompt.

### Deliberately not covered

**Coordinator tasks** don't get the badge. Sub-task propagation is fixed when the coordinator's MCP server registers, so turning the flag off there would stop the coordinator's own agent from skipping prompts while its sub-tasks kept skipping them until the next launch. For a safety setting, partly applying is worse than not offering it.

**Codex chat resume**: resuming a Codex thread after turning the flag off sends no `approvalPolicy`/`sandbox` on `thread/resume`, exactly as today, so the thread may keep the policy it started with. I couldn't verify whether Codex keeps it across a resume without a live Codex session. A new chat or a terminal restart applies the change either way; if you'd rather the resume send `on-request`/`workspace-write` explicitly when the flag is off, that's a small follow-up — I left it alone because it would also change resume behaviour for tasks whose flag never changed.

### Your call on the badge

With the default off, most tasks will show a muted "confirms on". Showing it only when on is a one-line change, but then turning it off from the title bar couldn't be undone there — and upstream's own chat note already tells users to "switch the task to skip permissions", which this gives them a place to do. Happy to go either way.

7 tests: the setter (on, off, no save when already off or absent, unknown task) and the Claude session reporting its mode on connect — which fails without that change. `npm run compile`, `check:static` and both suites pass.


miadisabelle:
Rebased onto 3.0.1 (`f190ba0`), no conflicts. One change for 3.x: delegation parents now also don't get the badge, alongside coordinators. Their children take `propagateSkipPermissions` (`delegation.ts`, `assignment.propagateSkipPermissions ?? task.propagateSkipPermissions`), which this badge doesn't change, so turning it off on a parent would only have stopped the parent itself — the same reason coordinators were excluded, and the same `!coordinatorMode && !delegationParent` pair the title bar already uses. The chat files are unchanged in 3.x, so that part stands as it was. `compile`, `check:static` and both suites pass.

